PR #7, opened 2019-03-22
BUG: day_night_background dates passed incorrectly to dates_between

AdamTheisen:
Dates passed to dates_between should be (start,end).  Current method was passing (end,start).


rcjackson:
CI has passed and there are no new routines so no new documentation is needed. I am merging.

PR #19, opened 2019-04-03
Clean up dataset from ARM QC to CF QC

kenkehoe:
I've added a class to call for cleaning up the dataset object from the native ARM QC to CF QC. I have it currently working as:
<pre>
import act
import act.io.clean

ceil_ds = act.io.armfiles.read_netcdf('/Users/kehoe/datastream/sgp/sgpceilC1.b1/*20180920*.nc')
ceil_ds.clean.cleanup()
</pre>
I'm a little confused why I need the _import act.io.clean_ part as I would have thought it would work with native import act, but it's not?  I think we could implement the read_netcdf to have the option to run cleanup() as well. You can also pass through **kwargs through cleanup() to the sub-methods to say what part of the cleanup should be implemented.

Also, I added a function to data_utils to extract missing value from attribute get_missing_value(). This would be used in cases like 
<pre>
ceil_ds = act.corrections.ceil.correct_ceil(ceil_ds,act.utils.data_utils.get_missing_value(ceil_ds,'backscatter'))
</pre>

Let me know what you think and if we can improve this. I also broke out the commits into different descriptions. I assume that is what we want.


act/utils/data_utils.py
```diff
@@ -53,3 +53,68 @@ def add_in_nan(time, data):
...
+            missing = data_object[variable].attrs[att]
+            in_object = True
+            break
+        except (AttributeError, KeyError):
+            missing = default
+
+    # Check data type and try to match missing_value to the data type of data
+    try:
+        missing = data_object[variable].data.dtype.type(missing)
+    except Exception as error:
```
rcjackson: It is generally not a good idea to do try/except statements unless you are catching specific errors instead of just skipping code for any error. I think it would be better to have the routine add the missing value attribute before this line with the name specified by the user if one is not found.

rcjackson:
I have a few comments here. First of all, in order to get "import act" to import all of the routines you need, you also have to edit the __init__.py in the io module. In there, you would have to add lines such as:

from .clean import cleanup

and so on for every function you want to make public in clean. 

In general, it is not good practice to use generic try/catch statements that will pass the code on any given error like this:
try:
    my_code = ""
except:
    pass

 (without specifying KeyError, AttributeError, etc.). So I would highly suggest rewriting those parts of the code where this is happening to remove those checks or make them catch specific kinds of errors. 

In addition, this module will need a unit test, which you can do by loading a small example dataset and testing to ensure that clean does what you expect it to do on that dataset. You would add such unit tests to the tests module of act. An example file that contains such tests is here:
 
https://github.com/ANL-DIGR/ACT/blob/master/act/tests/test_io.py

You would add your test to this module and run it using pytest. 

The code then needs to be brought up to PEP8 standards for style, but that we can worry about after the unit tests are developed and generic try/catch statements are removed.

rcjackson:
I pushed some final suggested style changes to your branch and fixed an error that popped up in another module that was caught by your unit test. (This is why we do unit testing!) Waiting for the CI to pass but all should be good to go.

PR #27, opened 2019-04-10
CSV Reader Addition Take 2

AdamTheisen:
I've updated my fork to the latest DIGR updates and added the csvfiles updates back in.  The CSV reader was built off of the csv files generated from data discovery.  Updates will likely need to be made when we work on other csv files.


AdamTheisen:
Added some additional code to be able to work with the ANL Tower data

rcjackson:
All tests have passed. The only thing that is needed now is a quick unit test for readcsv. Something simple where you check that certain outputs of the object are what you expect.

rcjackson:
I would put such a unit test in test_io.py.

AdamTheisen:
I have never done a unit test, but I will look at how you did things and see what I can do!  Thanks!

AdamTheisen:
@rcjackson I added in a unit test and made some additional updates to the CSV reader.  Let me know what you think.  Thanks!

AdamTheisen:
Nevermind, it was just showing green, but now says something failed. uploading fix

AdamTheisen:
@rcjackson The test that appears to be failing (test_multidataset_plot_tuple) does not run into the same error when I run on my laptop.  I didn't touch the areas where failing that I'm aware so I'm not quite sure what to do here.

AdamTheisen:
Line 281 in plotting.py is the only thing I could see causing this issue, but it works on my side so still not sure what's going on here

dsname = self._arm.keys()[0] -> dsname = list(self._arm.keys())[0]

rcjackson:
I traced it down to a bug in how ARMStandardsFlag was being initialized. Sometimes, it would set the default value of NO_DATASTREAM to True, and since it was not being set at the start, it could have been either True or False depending on what bit was previously there in memory. I've modified your armfiles.py and sent a commit to your master branch with the change.

AdamTheisen:
Thanks @rcjackson !

rcjackson:
More PEP8 fixes done. Also, I've set Travis to ignore warning 605 since you need that syntax for your unit test.

Issue #41, opened 2019-04-22
Plotting all NaN data causes error with y limits


kenkehoe:
File "/home/kehoe/dev/dq/lib/python/ACT/act/plotting/plot.py", line 365, in set_yrng
    self.axes[subplot_index].set_ylim(yrng)
  File "/apps/base/python3.6/lib/python3.6/site-packages/matplotlib/axes/_base.py", line 3616, in set_ylim
    bottom = self._validate_converted_limits(bottom, self.convert_yunits)
  File "/apps/base/python3.6/lib/python3.6/site-packages/matplotlib/axes/_base.py", line 3139, in _validate_converted_limits
    raise ValueError("Axis limits cannot be NaN or Inf")
ValueError: Axis limits cannot be NaN or Inf

We need to catch when all the data are set to NaN and set the y limits to something like [0, 1].

I can fix this if needed, let me know.

rcjackson:
Looks like this has been solved. I am closing the issue.

PR #42, opened 2019-04-23
Handlenan

kenkehoe:
This is fix issue #41 with handling when all data are NaNs.


rcjackson:
I am merging this in since all tests pass.

Issue #38, opened 2019-04-22
What to do when file not found


kenkehoe:
I would like to change the base behavior when a file is not found with io.read_netcdf() to catch the FileNotFound error and instead just return a value of None. This makes more logical sense to me to use the reading function to go check if the file exists. We do a lot of stuff depending on availability of files and having to wrap everything in a try seems excessive. I would suggest adding a verbose option to make a print statement optional if the file is not found, but not make that the default option. If this is OK I'll make the update.

AdamTheisen:
That sounds good to me

rcjackson:
@kenkehoe thank you for the PR fixing this issue! Closing.

PR #43, opened 2019-04-23
Returnnone

kenkehoe:
Adding **kwargs for xarray.open_mfdataset() and keyword option to return None when no files found. Should resolve issue #38 . Also updated for pep8.


rcjackson:
This looks like you're accidentially trying to overwrite plot.py and installation.rst. I would suggest making a new branch with just your new armfiles and csvfiles and trying again.

PR #44, opened 2019-04-24
Updated to return None when requested if no files found

kenkehoe:
My second attempt at updating process to read files and return None if no files found.


rcjackson:
All checks have passed. Merging.

PR #46, opened 2019-04-25
ADD: Skew-T plots and ability to mix plot types in subplot

rcjackson:
I have added Skew-T plots and the ability to mix differing plot types into one single subplot by linking together *Display objects. I hope this is of use.


rcjackson:
Tests have passed. @zssherman could you please review the code as a second pair of eyes?

zssherman:
@rcjackson Everything looks good to me. I was going to comment on having a try and except for metpy as optional, but I see it in the code, so your good there. Besides that I don't see anything pep8 related. For docs for documentation generation, do you need a space after the return list? I believe Py-ART's docs require it, but wasn't sure if it is needed in ACT. So like this:

Returns
----------
Blah : more blah
    more more blah

"""
If not then everything looks good to me. Maybe mention conda install for metpy on top of having the info on pip install.

PR #48, opened 2019-04-30
DOC: Updating contributor's guide.  More info on doc strings, PEP8, git branches and more.

zssherman:
@rcjackson Question, so the doc for contributors guide fails due to the python code blocks trying to import the actually library but I figured I save a space for the pull request, but is there a way to display python code in the rst file without using the actually act code?

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -162,27 +162,7 @@ Python File Setup
 -----------------
 
 When adding a new function to ACT, add the function in the __init__.py
-for the submodule so it can be included in the documentation. For example,
-adding get_date_format as a function from common module:
-
-.. code-block:: python
-
-        """
-        ===========================
-        act.plotting (act.plotting)
-        ===========================
-
-        .. currentmodule:: act.plotting
-        This module contains procedures for plotting ARM datasets.
-
-        .. autosummary::
-            :noindex:
-            :toctree: generated/
-
-            common.parse_ax
-            common.parse_ax_fig
-            common.get_date_format
-        """
+for the submodule so it can be included in the documentation.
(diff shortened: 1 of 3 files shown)


rcjackson:
I thought ..code-block :: python did not actually execute the code. What is your error?

zssherman:
@rcjackson:
Warning, treated as error:
[autosummary] failed to import 'TimeSeriesDisplay': no module named TimeSeriesDisplay
make: *** [html] Error 2

CONTRIBUTING.rst
```diff
@@ -54,51 +54,394 @@ If you are proposing a feature:
...
+
+        """
+        act.plotting.TimeSeriesDisplay
+        ==============================
+        Class for creating timeseries plots from ACT datasets.
+
+        .. autosummary::
+            :toctree: generated/
+
+            TimeSeriesDisplay
```
rcjackson: I would use a different doc string here? I think it's trying to import an RST file called TimeSeriesDisplay for some reason.

rcjackson:
I think it's looking for a module called TimeSeriesDisplay in the example docstring you have. This sounds like an issue we may need to raise with Sphinx autosummary. I would try using a different docstring example to see if that fixes the problem.

zssherman:
@rcjackson I just removed the code block because there kinda was already an explanation.

rcjackson:
Yeah I'm not sure why those aren't working. They should be, but I think it's a bug in sphinx in how it interprets doc strings in a code block. I'll commit for now and then we can work on a way to get the docs for __init__.py in. Perhaps we can just provide a hyperlink to an example __init__.py.

PR #52, opened 2019-05-02
Adding option to limit yrange when there are values outside a range

kenkehoe:
Adding option to limit yrange when there are values outside a reasonable range. Also fixed issue with yrng and multiple plots over a different range.


rcjackson:
All unit tests have passed and this would definitely be a useful fix. I'll merge this one in.

kenkehoe:
Thanks. I didn't look at the yrng issue with the other plotting classes. We should look into it if it could be an issue. Basically the yrng is being set on each data to be plotted, and the old range is discarded. Need to check if the old yrng should be used instead of new when setting new yrng.

PR #56, opened 2019-05-10
MPL Corrections

AdamTheisen:
Adding in a new MPL corrections script and test script. 

Fixing a bug in the plotting code where the cbmin and cbmax parameters were not being passed to pcolormesh


rcjackson:
The unit test is failing at the obj.where statement. I wonder if the data being read in has non-numerical values going by the error.

AdamTheisen:
Paytsar found a bug and will push that fix shortly.  I will look into the test and why it's failing this morning as well.  It didn't throw any errors in testing on my system, but there could very well be some NANs in the data.  That will be expected with this dataset.  I'll try and clean up the data to make sure there aren't any issues for the test.

rcjackson:
Looks good to me. Merging.

PR #58, opened 2019-05-13
Geoplot2

kenkehoe:
I was having an issue with the test running and test for plot_from_u_and_v(). I also made a few minor changes to the new plotting stuff I did. This is not fixing my py.test issue but I do need the plot_from_u_and_v() to let the test pass.

### DIFF
--- a/act/plotting/plot.py
+++ b/act/plotting/plot.py
@@ -1520,7 +1520,7 @@ class SkewTDisplay(Display):
         v_red = np.zeros_like(p_levels_to_plot) * getattr(units, v_units)
 
         for i in range(len(p_levels_to_plot)):
-            index = np.argmin(np.abs(p_levels_to_plot[i] - p))
+            index = np.argmin(np.abs(p_levels_to_plot[i] - p.magnitude))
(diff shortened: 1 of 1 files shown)


AdamTheisen:
Requesting Bobby as a reviewer.

I never had issues with the plot_from_u_and_v test so I am a little concerned about potential conflicts on the ADC system.

rcjackson:
I wonder if there is some sort of setting in the ADC environment that is making it fail. I'll try the test on my own end and see what I get.

rcjackson:
For your py.test issue I tried simply running the test manually and I get this figure.
myfig

Therefore, I think there might be an issue with pytest-mpl when it preserves the figure colormap. I'll look into raising an issue on their webpage.

rcjackson:
I have raised this as an issue. We can see if it's truly a problem on their end or something we need to do.

https://github.com/matplotlib/pytest-mpl/issues/84

kenkehoe:
OK thanks. Can we comment out the test with our stuff for now?

rcjackson:
Yeah let's not do the test for now.

Bobby

PR #63, opened 2019-05-20
Clean state

kenkehoe:
This is an update to help fix issues with xarray converting any variable with _FillValue or missing_value attribute set.  If xarray finds a variable like this it will convert the variable to float and change any missing value indicator to nan. This is fine for data but not good for qc or state fields.

I also updated some of the other functions as there were some minor issues and duplication.


AdamTheisen:
@kenkehoe  Looks like just one pep8 issue


./act/io/clean.py:129:17: W503 line break before binary operator

Issue #64, opened 2019-05-21
read_netcdf function causes segfault when attempting to read in a day of radar data


atkingtornado:
It seems that the read_netcdf function in act.io.armfiles causes a segfault when reading in radar data. Perhaps because of the large file sizes? the specific case that causes the segfault is reading in the following list of files:

[...]

PR #65, opened 2019-05-21
Change open_mfdataset parallel to keyword default to False

AdamTheisen:
Also including a change to description in setup.py


AdamTheisen:
Note, this will address issue #64

kenkehoe:
I think Corey's pull request is better since it allows for user to control all the keywords. Just changing the True to False will not allow a user to decide.

AdamTheisen:
Going with Corey's instead

PR #66, opened 2019-05-21
changing arguments in reader

cgodine:
The DQ office was having issues reading multiple radar files, removing the parallel argument solved this issue. We believe that removing the keyword and having the option to pass in parallel=True explicitly is a better option. I also removed the variables keyword since it was not being used, and moved the concat_dim keyword into the function call.


AdamTheisen:
Let's go with this over the pull i requested

PR #68, opened 2019-05-23
Fixing issue with cleaning up global attributes but there are no glob…

kenkehoe:
Fixing issue with cleaning up global attributes but there are no global attributes to clean.


AdamTheisen:
Looks to be a good fix

Issue #55, opened 2019-05-08
Stability Indices for BBSS data 


Trekev:
It may be useful to have a way to calculate stability indices for the BBSS data.  Perhaps a way to output a table with the plot or a separate file.

rcjackson:
I think this would probably go under the "retrieval" module. I envision a first entry in that module with "calc_stability_indices" that returns a dictionary with CAPE, CIN, LI, etc.

scollis:
Very cool.. Easy hook to MetPy

PR #70, opened 2019-06-03
First retrieval in ACT: Compute CAPE, CIN, LCL, LFC, and lifted index from sounding.

rcjackson:
ACT finally has a retrieval in it! Now we can compute CAPE, CIN, LCL, LFC, and lifted index from a sounding. This addresses #55.

### DIFF
--- a/act/__init__.py
+++ b/act/__init__.py
@@ -4,6 +4,7 @@ from . import corrections
 from . import utils
 from . import tests
 from . import discovery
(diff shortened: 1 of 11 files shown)


AdamTheisen:
Looks like maybe a file missing from repo?

[autosummary] failed to import 'act.io.coordinates.assign_coordinates': no module named act.io.coordinates.assign_coordinates

PR #71, opened 2019-06-04
ADD: Small fix to TimeSeriesDisplay

rcjackson:
This fixes an error when trying to plot barb time series from two dimensional u, v fields.


rcjackson:
@scollis the deed is done.

PR #73, opened 2019-06-04
STY: Multiple style changes, mispelling, imports and more.

zssherman:
To be more consistent between submodules and code, added colon between parameter definitions in docs. Fixed a few spelling errors and added space at the end of doc strings to be more in line with nump standards. I also changed the import of the utils to remove the need for a longer extension. IO import are the same to keep different datatype modules separate.

### DIFF
--- a/act/corrections/__init__.py
+++ b/act/corrections/__init__.py
@@ -12,6 +12,7 @@ The procedures in this module contain corrections for various ARM datasets.
 
     ceil.correct_ceil
 """
+
 from . import common
 from . import ceil
 from . import mpl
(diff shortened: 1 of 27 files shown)


AdamTheisen:
@zssherman Looks like this one failed.

Warning, treated as error:
/home/travis/build/ANL-DIGR/ACT/docs/source/API/generated/act.utils.data_utils.add_in_nan.rst:document isn't included in any toctree
make: *** [html] Error 2

zssherman:
Hmm yeah the act.utils.data_utils.add_in_nan path shouldn't exist anymore with my changes so when docs does make html it should regenerate the new files unless I'm mistaken. But that's why I'm slightly confused on whats going on. @rcjackson your thoughts?

rcjackson:
Try removing everything in your ACT/docs/source/API/generated and commit. That directory is built after Sphinx makes the docs and is not needed. Sometimes this can cause conflicts with the current build.


Bobby

PR #79, opened 2019-06-27
adding day night background for 24 hour sunlight

cgodine:
When the day night background is used in plotting (specifically at high latitude sites such as NSA), astral will throw an error when the sun doesn't set in the summer months. I made a small update that will make the plot continuously day-colored if the sun doesn't set below the six degree horizon threshold set by astral.


rcjackson:
Tests passed. This looks ok to me.

PR #84, opened 2019-07-11
Persistence test

kenkehoe:
I'm adding a persistence test to the qctests module.

While making this I discovered that the clean module had an issue setting flag_masks and flag_assessments. Instead of being a copy was a direct link. Had to update to use a copy.copy.

Also found an issue setting xrange with plots when only one value to plot. It returns a warning. Capturing the warning.


AdamTheisen:
@kenkehoe it looks like whatever you put in for xrange is causing the plotting tests to fail.

rcjackson:
It's failing at:
with warnings.filterwarnings("ignore"):

I don't think you need the with. I've always just done "warnings.filterwarnings("ignore")."

rcjackson:
Looks like that fixed the issue. I am merging.

kenkehoe:
Thanks. Do we have a policy on catching warnings? I didn't keep it wrapped in a with since I can use a regular expression to catch the individual warning. I think we should encourage doing something like that instead of catching all warnings. I'm also not sure how long that warning catch will propagate throughout the codes so keeping a warnings catch to individual warnings would be a good practice.

AdamTheisen:
good question.  @rcjackson or @zssherman do you have any general guidance?

zssherman:
@kenkehoe @AdamTheisen It depends, I usually do use:
```with warnings.filterwarnings("ingore", FutureWarning)```
for when an individual line is throwing a warning that can be ignored for only that line. Sometimes ignoring all of one warning can be useful, idk I personally try to use with warnings because sometimes catching depreciation warnings is key to update code before a dependency breaks it.

PR #85, opened 2019-07-12
Adding a difference test for time series data. 

kenkehoe:
Adding a difference test for time series data.  Also adding a function to convert units on data arrays wrapped around pint.

### DIFF
--- a/act/qc/qcfilter.py
+++ b/act/qc/qcfilter.py
@@ -243,7 +243,7 @@ class QCFilter(object):
                  test_meaning=None, test_assessment='Bad',
                  flag_value=False):
         '''
-        Method to add a new test/filter to a qualit control variable.
+        Method to add a new test/filter to a quality control variable.
 
         Parameters
         ----------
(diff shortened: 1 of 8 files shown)


AdamTheisen:
@kenkehoe one small pep8 issue

/act/qc/qctests.py:947:1: W293 blank line contains whitespace

AdamTheisen:
@rcjackson @zssherman it looks like this will make Pint a dependency.  I am fine with this, but wanted to get your thoughts as well.

zssherman:
@AdamTheisen I think its fine, the unit functionality in pint might come in handy down the road for other functions etc.

kenkehoe:
OK I'll fix the error. I'm a little confused about pint dependency. I'm pretty sure "units" used elsewhere is actually using pint. Does a different library using pint not count as using the pint dependency?

zssherman:
conda will install dependencies of dependencies, so it might be installing pint already due to xarray or dask. Think the only way it would make a difference is if you wanted to use pint, but a dependency for some reason got rid of it. But not sure the best route, you might be able to leave it out of the yml files for now.

PR #88, opened 2019-07-19
Catch time_offset error

kenkehoe:
When I added the ability to read older ARM files that did not have time variable it added a bug where the options to tell xarray to not parse as CF file is causing an error. I've added an exception to allow reading the netCDF file in its raw form in case that is important.


kenkehoe:
I found out how to have methods to be used under the same class name can be declared in different files. I implemented this in the last push.

PR #89, opened 2019-07-22
Updating to include QC in the documentation.  Also changing the plot examples that use metpy to regular examples until bug is resolved in metpy

AdamTheisen:


rcjackson:
This looks like the correct way to include the qc module in the documents. The build passes, so I will merge and it should appear on the page now.

PR #90, opened 2019-07-22
High lat background for plots

kenkehoe:
The exception to catch errors when plotting at high latitude and no sunset is not working correctly. I've updated the exception to use plot range instead of trying to get time range from object. Also, updated the exception to be more specific.

Added exception to catch errors when location variables are not in the object. It will do a smart search for known variable names first, then look for standard_name then return gracefully.

Also, minor issue with cleanup and checking long_name.

### DIFF
--- a/act/io/clean.py
+++ b/act/io/clean.py
@@ -455,9 +455,12 @@ class CleanDataset(object):
             except AttributeError:
                 continue
             # Skip data quality fields.
-            if not ('Quality check results on field:' in
-                    self._obj[var].attrs['long_name']):
-                continue
+            try:
+                if not ('Quality check results on field:' in
+                        self._obj[var].attrs['long_name']):
+                    continue
+            except KeyError:
+                pass
 
             # Get existing data variable ancillary_variables attribute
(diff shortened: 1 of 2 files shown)


rcjackson:
This all looks good to me. What do you think @AdamTheisen ?

kenkehoe:
OK, I fixed an error that i contacted @AdamTheisen about in Slack I thought of while having dinner at Sherpa's. In the process of fixing that found a bunch of other fun stuff. 

1) Need to loop over previous, current and next day to get all the times when sun is up. 
2) astral also has a strange way to calculate sunrise and sunset where it throws an error if sun never gets below 6 degrees below horizon. Had to set that to 0 get the jump from some night to no night to be closer to smooth. Still not 100% but not sure how to fix.
3) Because error for all day and all night is the same need to check solar azimuth at noon to see if the times are for all day or all night. 

Overall much better but someday we can try to get the transition perfect.

AdamTheisen:
This was definitely needed, thanks @kenkehoe

PR #91, opened 2019-07-25
Adding Ability to Query CropScape API to get Crop Type

AdamTheisen:


AdamTheisen:
Removing Bobby as reviewer and adding Zach.

zssherman:
@AdamTheisen For the code itself, I would maybe move the imports outside of the functions, personal preference I guess, but ultimately up to you.
Documentation wise, I would add for the function the category return in the doc string:
returns
---------
category : etc
    A etc etc
There are also some extra spaces in the docstring paragraph too.
Besides that looks good to me.

zssherman:
Oh and import order for pep8 should be:
import datetime

import pyproj
import requests

datetime is python built in package and pyproj and requests are third party

AdamTheisen:
thanks @zssherman  I made the changes and added a test as well.

AdamTheisen:
@zssherman what do you think?  good to go now?

zssherman:
@AdamTheisen Yeah, I would say its ready to merge.

PR #94, opened 2019-07-29
Fixing issue I introduced with recent update detecting QC variables.

kenkehoe:
Fixing issue with "(" and ")"  in re after switching to new method for detecting QC variables.

I ran the pytest -v locally and there is a failure in something I didn't touch. Not sure if this will have a failure or not.


AdamTheisen:
Looks like a minor update.  will merge in since all checks passed.

PR #97, opened 2019-08-02
ENH: Updating read_netcdf open_mfdatastet to fix xarray deprecation errors

AdamTheisen:


zssherman:
@AdamTheisen So the default combine parameter choice I believe was auto_combine before the new warning. Is nested return the same result as the auto_combine?

AdamTheisen:
Yes.  auto_combine is being deprecated.  FutureWarning: Also `open_mfdataset` will no longer accept a `concat_dim` argument.
To get equivalent behaviour from now on please use the new
`combine_nested` function instead (or the `combine='nested'` option to
`open_mfdataset`).

The other option is to pass combine='by_coords' to open_mdfdataset, but that results in errors.  It seems like the may I have is the appropriate method but am open to suggestions

zssherman:
@AdamTheisen Looks good to me then, I missed the 'please use the new combine_nested' part lol

PR #100, opened 2019-08-08
Return qc

kenkehoe:
Adding option to return inverted masked array for indicating where test is not set. This will be used to overplot where failing. Also fixed an issue with additional description values for determining QC variables.


kenkehoe:
Don't merge yet. I want to see if I can fix the setting xrange to single time warning again...

kenkehoe:
OK I have updated the with statement in plotting to catch more general warning messages. Hopefully this fixes the issue.

AdamTheisen:
Everything looks good and has passed testing.  Will merge in.

PR #105, opened 2019-08-22
Trying to get qc documentation to render correctly. 

kenkehoe:
Made some fixes but still not working.

Please hold off on merging.

### DIFF
--- a/act/qc/__init__.py
+++ b/act/qc/__init__.py
@@ -11,8 +11,12 @@ and returning a mask from the quality control variable.
 .. autosummary::
     :toctree: generated/
 
-    qcfilter
-    qctests
+    qcfilter.QCFilter
+    qctests.QCTests
+
 """
 
 from . import qcfilter
+from . import qctests
+# from .qcfilter import QCFilter
+# from .qctests import QCTests
--- a/act/qc/qcfilter.py
+++ b/act/qc/qcfilter.py
@@ -1,6 +1,8 @@
 """
-act.qc.filter
-=============
+===============
+act.qc.qcfilter
+===============
+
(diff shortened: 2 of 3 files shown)


kenkehoe:
I'm only doing the pull request so you can see where I'm at. I fixed some issues with document strings but it still does not work. I keep getting an error I don't understand.

Warning, treated as error:
/Users/kehoe/dev/dq/lib/python/ACT/docs/source/API/generated/act.qc.qcfilter.rst:document isn't included in any toctree

Anyone have suggestions?

AdamTheisen:
There's a class defined in qcfilter of QCFilter.

kenkehoe:
I've tried every possible permutation. I don't know.

zssherman:
@kenkehoe For qcfilter, the path is act.qc.filter, when it should be act.qc.qcfilter to match the module path

zssherman:
Nvm see it as part of your changes

rcjackson:
The CI passed, so the docs are not failing to build from a clean build, but only on Ken's machine. 

On your machine, Ken, try demoving every file in the /Users/kehoe/dev/dq/lib/python/ACT/docs/source/API/generated/ directory. This is only generated after the make is done and if there are unused files there it will fail.

AdamTheisen:
I think I have a fix for all this based on Ken's updates.  If you want I can do a pull request.  It would also include the new cloud base height detection algorithm I have.

kenkehoe:
If @AdamTheisen  has a possible solution then I suggest we try that.

AdamTheisen:
@kenkehoe I'll close this pull request and then have you submit new once you update the code from #107

PR #114, opened 2019-09-06
Updates to MANIFEST for anaconda distribution.

AdamTheisen:
This adds the requirements files to the MANIFEST for builds.  There were also some issues with the plotting tests that I resolved.  Took out plt.show() since the test_plotting was using the AGG backend.


AdamTheisen:
Will merge and do a microversion update

PR #115, opened 2019-09-06
Additional MANIFEST Inclusions

AdamTheisen:
The new plotting txt file was left out.


rcjackson:
I think we're good now.

PR #120, opened 2019-09-11
Fixing warning caused by looking for a color "red" in a non-dict object.

kenkehoe:
After the color map update a new warning was displayed indicating issue looking for 'red' in a non-dictionary object. I'm adding an if statement to make the warnings go away.


AdamTheisen:
Checks passed, thanks for this fix @kenkehoe!

PR #122, opened 2019-09-17
DOC: Added Anaconda badges and added act-atmos conda-forge install info.

zssherman:


AdamTheisen:
Thanks for adding this in @zssherman

PR #126, opened 2019-09-25
Clean fix

kenkehoe:
Fixing a few issues with the cleanup method, specifically with updating from ARM state to CF state variables. Also, found out how to stop one of the warnings with testing.


AdamTheisen:
Thanks @kenkehoe.  This looks good

PR #127, opened 2019-09-28
ADD: New functions for correcting wind direction and speed for ship m…

AdamTheisen:
Functions for calculating course and speed over ground and for correcting wind data for ship motion.  Calculations have been tested over a months worth of MAR NAV and AOSMET data and compared with the MAR SONDE wind direction and speed data.

ship_corrections


rcjackson:
I would write a unit test for this entry since it's a new feature, but otherwise everything else looks good.

zssherman:
@AdamTheisen I approved, but I agree with Bobby, just unit tests, but everything else looks good.

AdamTheisen:
Adding in 2 tests for course over ground calc and wind corrections.  Also adding in an example and fixed some documentation issues I came across while adding in the ship calcs to the autodocs.

AdamTheisen:
Sorry, my flake8 did not catch those pep8 issues.

PR #129, opened 2019-10-01
ADD: Ability to plot size spectra.

rcjackson:
In my SP2 data processing, I thought it would be useful to have ACT support the plotting of size distributions, since this can also be used for many other instruments as well (CIP, 2DC, etc.). I simply added a new method to HistogramDisplay.


act/tests/test_plotting.py
```diff
@@ -257,6 +258,21 @@ def test_heatmap():
...
+@pytest.mark.mpl_image_compare(tolerance=30)
+def test_size_distribution():
+    sigma = 10
+    mu = 50
+    bins = np.linspace(0, 100, 50)
+    ydata = 1 / (sigma * np.sqrt(2 * np.pi)) * np.exp(-(bins - mu)**2 / (2 * sigma**2))
+    y_array = xr.DataArray(ydata, dims={'bins': bins})
+    bins = xr.DataArray(bins, dims={'bins': bins})
+    my_fake_ds = xr.Dataset({'bins': bins, 'ydata': y_array})
+    print(my_fake_ds)
```
AdamTheisen: Will this print statement affect the test at all?

AdamTheisen:
@rcjackson while the tests are running, do you have an example plot you could upload here?  I'm just curious

rcjackson:
example_wave

Example scattering signal from SP2.

PR #134, opened 2019-10-10
Test update

kenkehoe:
Fixing some issues with persistence test and instrument comparison tests. Also now allowing None to be set for limits to indicate test should not be run. Something the DQ Office sets up in a loop.


AdamTheisen:
All checks pass and looks like good updates. merging

PR #135, opened 2019-10-11
Create OBJ from ARM DOD

AdamTheisen:
Adding function and corresponding test for creating an object from an ARM dod through the PCM API.  Testing on this has included creating ACT object, assigning data to object, writing to netcdf, and then reading back in and plotting.

Also fixed an error with TimeSeriesDisplay when units are not available in the ydata dataarray

### DIFF
--- a/act/io/__init__.py
+++ b/act/io/__init__.py
@@ -13,6 +13,7 @@ This module contains procedures for reading and writing various ARM datasets.
     armfiles.read_netcdf
     armfiles.check_arm_standards
     armfiles.ARMStandardsFlag
+    armfiles.create_obj_from_arm_dod
     dataset.ACTAccessor
     csvfiles.read_csv
 """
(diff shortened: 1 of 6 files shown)


AdamTheisen:
Good call @rcjackson .  I see that now in the return information and will add in.  Shouldn't be too much trouble to switch things around.  Will probably still need a number of time samples passed in, but it looks like I should be able to grab it for the other dims.  Will default to the DOD, then if the user passes in an overwrite will use that.

AdamTheisen:
Update pushed.  If the dimension size is note passed in, it will look to the DOD.  If not in the DOD, then will throw an error.  This will allow users to set different sizes for dimensions if needed.  Note, sending the number of time samples is still required though (i.e. {'time': 1440}

rcjackson:
Awesome. I don't have any further comments myself. Thank you for doing this!

PR #136, opened 2019-10-11
Adding option to use colormap with wind barb plot. Also fixing issue …

kenkehoe:
…with day_night_background overplotting data.


AdamTheisen:
Looks good, thanks for doing this!  Adding in the image you sent me for reference.

corsodarM1 b1 sodar_winds 20190429

PR #137, opened 2019-10-14
Addition of geo_utils

AdamTheisen:
Adding a function to utils area to calculate the destination lat/lon given a starting lat/lon and azimuth/distance.  There are libraries out there that do this, but was hesitant to add dependencies and given the simple nature thought it would be a good addition.


zssherman:
Looks all good to me.

PR #141, opened 2019-10-16
ENH: Documentation Improvement

AdamTheisen:
Fixing the documentation of the parameters and returns for almost all programs.  Changed the ' : ' to ': ' between the variable name and the type.  Also updating the documentation for the wind retrieval algorithm from DL PPIs to properly credit where the code came from.  Finally, updated the wind ship corrections to include proper reference to NOAA document.


zssherman:
@AdamTheisen Looks good to me. I'm use to there being a space between both colons, but sphinx generation seems to be different for ACT then Py-ART so the standards must have change, but overall I think its good to go.

PR #142, opened 2019-10-16
Dl winds3

kenkehoe:
Fixing a bug with calculation of wind direction, adding some additional variables in the returned object that are used by mentor and maybe users.


AdamTheisen:
Good catches on these.

PR #143, opened 2019-10-24
Fixing issue with add_in_nan when there is only one time step.

kenkehoe:
When there is only one time step add_in_nan has a problem with the mode stats.


AdamTheisen:
Error is from a MetPy upgrade that affected the values we were checking in the stability test.  Working to resolve in #144

AdamTheisen:
This looks good to go now.  Merging

PR #146, opened 2019-11-05
Moving ACT attributes to the xarray attrs

AdamTheisen:
Moved the obj.act attributes to standard obj.attrs.  I've updated the plotting code as well.  As part of this I removed some of the attributes that were duplicated in ob.attrs with armfiles.py.  Tests seem to all be passing on my side, but just want everyone to take a look and see if this is an ok update to address #8 .  @kenkehoe thoughts?


AdamTheisen:
Good call @rcjackson  I removed it and worked through the bugs that came up in the tests.  Hopefully this will simplify things a bit

AdamTheisen:
If @kenkehoe is good with this as well I will go ahead and merge.

kenkehoe:
Sounds good.

PR #147, opened 2019-11-18
ADD: Adding method to calculate active AOSPURGE times.

cgodine:
When there is a high amount of contaminant in the AOS, the purge system will activate and flush the contaminated air. During this time the instrument is not sampling ambient air and those time ranges should be noted in a DQR. This script will take in an xarray Dataset of AOSPURGE data as read by act.io.armfiles.read_netcdf and return a list of time ranges when the purge system is active. It also contains an option to save the time ranges to a .txt file that can be read into the DQR submission tool.

### DIFF
--- a/act/retrievals/__init__.py
+++ b/act/retrievals/__init__.py
@@ -14,9 +14,11 @@ This module contains various retrievals for ARM datsets.
     generic_sobel_cbh
     calculate_precipitable_water
     compute_winds_from_ppi
+    calculate_aospurge_times
 """
 
 from .stability_indices import calculate_stability_indicies
 from .cbh import generic_sobel_cbh
 from .pwv_calc import calculate_precipitable_water
 from .doppler_lidar import compute_winds_from_ppi
+from .purge_times import calculate_aospurge_times
--- a/act/retrievals/purge_times.py
+++ b/act/retrievals/purge_times.py
@@ -1,0 +1,107 @@
+import os
+
+import numpy as np
+import pandas as pd
+
+
+def calculate_aospurge_times(obj, threshold=None,
+                             txt_path=None):
+
+    """
+    Method to calculate times where the AOS purge system is active. It will
+    print out the start and end times and has the option to save those time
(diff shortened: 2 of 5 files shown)


AdamTheisen:
@cgodine there's an error with the test

    def test_calculate_aospurge_times():
        purge_ds = act.io.armfiles.read_netcdf(
            act.tests.sample_files.EXAMPLE_AOSPURGE)
>       purge_times = act.retrievals.purge_times.calculate_aospurge_times(filename=purge_ds)
E       TypeError: calculate_aospurge_times() got an unexpected keyword argument 'filename'

However, I think we should discuss if there's a different way we could approach this, @kenkehoe @scollis @rcjackson @zssherman .  This function is very specific to the ARM AOS Purge System, but I think with a little bit of tooling, you could make a more generic function out of it that would benefit a wide range of applications.  

I.e. a general function to spit out start/end times of any time series given a lower/upper threshold.  Thoughts?

cgodine:
@AdamTheisen I'm actually in the process of making a generic method that does this. I have a few more things I'd like to add to it but I have code committed to our dqlib library under retrievals/dqr.py.

AdamTheisen:
Additional PEP8 issues:

./act/tests/test_retrievals.py:62:1: E302 expected 2 blank lines, found 1
./act/tests/test_retrievals.py:66:75: W291 trailing whitespace
./act/tests/test_retrievals.py:67:75: W291 trailing whitespace
./act/tests/test_retrievals.py:68:75: W291 trailing whitespace

AdamTheisen:
I'll wait to hear what others think with regards to this.  It might make more sense from the ACT side of things to wait for that generic code, but am open to other thoughts/suggestions.

AdamTheisen:
@cgodine We discussed on the ACT call today and agreed this would be better in the DQ python libs since it is very specific to the AOSPURGE system.  We also agreed that the generic version you are working on in python would be a valuable addition and look forward to that PR.

PR #149, opened 2019-11-23
Dl update

kenkehoe:
- Update to the DL corrections to perform range normalization correctly.
- Updates to DL retrieval to make looping over many scans easier.
- Added test for DL correction
- Fixing bug with TimeSeriesDisplay wind barb plot. With my initial commits I was confused with how to set number of points to plot for x and y. They seemed to be reversed with the keywords. But I just thought I was doing it wrong. After looking at it again I am sure they are reversed. So now the x and y should be correct with keywords. That one caused some loss of brain cells to diagnose.

I tried to merge current master with this branch before creating PR. Hopefully this all works out correctly.

### DIFF
--- a/act/corrections/doppler_lidar.py
+++ b/act/corrections/doppler_lidar.py
@@ -1,11 +1,14 @@
 import numpy as np
+import pint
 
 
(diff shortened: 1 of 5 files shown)


AdamTheisen:
This all looks good, will merge.

PR #150, opened 2019-12-02
Adding a Try to catch when file_dates not set in object.

kenkehoe:
I'm getting an error while plotting data not read in with the ACT reader. This try will allow data to be plotted with an Xarray Dataset read with Xarray only reader. I think this is the last bit of removing the act object variables.


AdamTheisen:
Good catch!  Thanks @kenkehoe

kenkehoe:
Can we generate a new version so I can get this installed on ARM dev machines.

AdamTheisen:
It looks like it's been a while.  Will draft up a new one.

PR #151, opened 2019-12-03
ADD: Adding method to calculate times of bad/missing data to be used …

cgodine:
…for DQRs.

This is a method to calculate time ranges of bad or missing data, the time stamps of which can be used for the DQR submission tool. This is a more generalized version of the code in PR #147 which focused solely on AOSPURGE beginning and end times.


AdamTheisen:
@cgodine  it looks like the AOSPURGE code was added into this as somewhat of a standalone section.  Is there a different way to approach that in the DQ processing?  i.e.

Add QC tests to object flagging when the aospurge is 1 (DQ Processing)
Use this dqr_times test to get the times when that test is tripped

Thoughts on that? I think Ken had it set up to easily add tests to an object.

I would also recommend that these functions and corresponding unit test be better suited in the qc modules as it is working off of qc bits.  @kenkehoe thoughts on that?  utils would work was well if you wanted to create a qc_utils.py file under it.

cgodine:
Hi Adam,
Thank you for the feedback! I was a little unsure of how to approach the AOSPURGE data since it behaves differently than other QC, but adding a QC test to the data and using the existing code seems like a better way to do it. I'll make changes to that. 

I'll also move everything over to qc. Making an additional test is quite simple. I think you're right in that the methods would be better suited there, especially with an additional QC test being written for them.

cgodine:
After going through and simplifying the code a little more, I decided to move the methods into the utils directory under a qc_utils.py file. Seems to be a more appropriate area for them.

AdamTheisen:
@cgodine This is looking good!  I just have a question on one part of the functionality.  Am i understanding it correctly that it will only process a single date at a time or will it be able to work with a range of files?

cgodine:
It will work for a range of files so long as they are combined into one xarray object that gets passed to the function. Here is the output from a week's worth of aospurge data:
Screen Shot 2019-12-04 at 12 40 36 PM

AdamTheisen:
Thanks @cgodine !  I think this can be merged, I do just have one last request.  We are trying to keep a limit on the size of data files in the test area.  The aospurge one would actually be the second largest one in there.  Can we get by in the test without testing AOSPURGE?  Or if you still wanted to included it, could you trim the file size down by subsetting the data.  I.e. for the MPL data, I had just dumped a short time period out to netcdf using xarray.

cgodine:
@AdamTheisen I switched out the AOSPURGE test data and added a smaller netCDF file. Everything should be running smoothly again!

AdamTheisen:
@rcjackson or @zssherman Are you able to go in and restart the build?  I'm not sure what's going on here and I don't see to have access anymore to restart the build.

PR #153, opened 2019-12-06
DOC: Removing odd statement in the installation section.

rcjackson:
I found an outdated statement in the documentation saying that we soon "plan on implementing pip and conda install" when we did that ages ago. I have removed the statement from the docs.


rcjackson:
Minor edits to docs, merging.

PR #154, opened 2019-12-06
Time fix

kenkehoe:
We had an issue where the units of the time variable were not set correctly but base_time and time_offset are in the file. Using those we can fix the issue. This PR is to implement that fix.


AdamTheisen:
Looks good and all checks pass.  Adding in

PR #155, opened 2019-12-09
QC change

kenkehoe:
Austin discovered an issue with QC when the location of QC bit description changed from global to variable attributes. Turns out that open_mfdataset() with combine='nested' will just not read any attribute that changes between read data files. This results in a problem parsing the attributes for determining QC information. After looking into options I discovered that setting combine='by_coords' will read the attributes and keep the ones set by last data file. This allows us to get something to work with and then fix it once read in. This should also do some more magic to get the ordering of values in a coordinate to be monotonic. I think it will be better overall. I initially put in these keywords to stop xarray from printing a warning.

I had to modify the one test code to use 'nested' when it had a dimension not in the file 'time'. To enable users to set all the values all options sent to open_mfdataset() are now setable with the read_netcdf() call.

Also found use of enumerate() in a for loop where it was not being used.

### DIFF
--- a/act/io/armfiles.py
+++ b/act/io/armfiles.py
@@ -45,26 +45,33 @@ class ARMStandardsFlag(Flag):
     """ The dataset does not have a datastream field. """
 
 
-def read_netcdf(filenames, concat_dim='time', return_None=False, **kwargs):
+def read_netcdf(filenames, concat_dim='time', return_None=False,
+                combine='by_coords', **kwargs):
     """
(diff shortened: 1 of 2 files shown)


AdamTheisen:
@kenkehoe  Let me test this out a bit.  I remember there being some reason that I went with nested over combine.  I want to run some of my test code and see if there are any failures first before proceeding.

AdamTheisen:
I've run a number of tests on my end and all look to work with this version.  Will go ahead and merge

Issue #156, opened 2019-12-10
Adding internal global attributes to Xarrray Dataset to replace ACT object attributes


kenkehoe:
Initially ACT would add some attributes to the Xarray Dataset that are used by the plotting functions. Some of them include file_dates, file_times, and datastream. We changed to adding them as Dataset global attributes so others who don't use our reader can fix the missing information when needed.

But the part I'm not a fan of is adding these without an indication they are part of ACT not part of the data file. I suggest prepending "__" to any internal information added to the Dataset so we know and can easily remove them if needed. The current variables I see file_dates, file_times, arm_standards_flag and datastream should be changed to __fiile_dates, __file_times, __arm_standards_flag and __datastream.

This is also causing issues with the variables that do exist in the data file like "datastream". We have some code that fixes issues when processing older ARM data that are missing global attributes, particularly datastream. Since ACT is adding that global attribute our code thinks things are OK and we have other issues.

Here is an example of a Dataset global attriubutes:
```
    ...
    serial_number:              P2710385
    qc_standards_version:       1.0
    qc_method:                  Standard Mentor QC
    qc_comment:                 The QC field values are a bit packed represen...
    zeb_platform:               enasondewnpnC1.b1
    history:                    created by user dsmgr on machine garnet at 20...
    file_dates:                 ['20191206']
    file_times:                 ['113100']
    datastream:                 act_datastream
    arm_standards_flag:         ARMStandardsFlag.OK
    __source_files:             ['enasondewnpnC1.b1.20191206.113100.cdf']
```

Everything after history: is added by ACT or DQ Office code. We add __source_files.

AdamTheisen:
This sounds like a good idea. @rcjackson @zssherman any objections?

@kenkehoe do you know offhand how much effort this would be to change up?

kenkehoe:
I don't think our code use any of these. Or if we do they are rare. I would assume the updates would be minimal.

AdamTheisen:
I think I found all the instances and have it changed on my local code.  Running tests now to ensure I didn't miss anythign

AdamTheisen:
@kenkehoe check out the PR indicated above and see if that was what you were thinking

kenkehoe:
yeah those look correct. Do we want one or two underscores at the start. I'm currently using double underscore but looking at PEP8 I think single is what we want. I'll update our code to match.

rcjackson:
This is fine with me!

PR #157, opened 2019-12-10
ENH: Updates to better define the attributes that are added as part o…

AdamTheisen:
…f ACT.  Includes corresponding changes to the unit tests as well.  Also, in running flake8 on the code, it indicated that a number of imports were not being used in the plotting so I removed them.

Example for discussion in #156


AdamTheisen:
@rcjackson @zssherman any issues with changing to this method for storing attributes added by ACT?

zssherman:
@AdamTheisen Fine with me, usually underscore is private variables etc, but I think in this case its desired, so I don't see an issue.

PR #158, opened 2019-12-12
A few changes to fix how xarray reads the time units string. 

kenkehoe:
Also updated the error handling for FileNotFoundError.

This all came about after we updated combine='by_coords'. I still think this is the correct method but we discovered that the way Xarray reads the units string with time is not correct. It does not handle the timezone offset part at the end of the units string. This was causing it to not set the time correctly, but date was correct. To fix this we are setting use_cftime=True to force it to use the cftime library to fix the units string reading part. The issue is that this will read the data in as cftime and leave it that way in the returned object. The second part of this is to convert the time back to expected numpy datetime64 values.

I also updated the way we handle the errors so we can fine tune the error handling and use the error string to parse specific errors and try to handle them. If we get a very specific error we will change combine = 'nested' and try again. If it still fails or fails for a different ValueError reason will just throw that error right away.

A lot of changes but all for the better.

### DIFF
--- a/act/io/armfiles.py
+++ b/act/io/armfiles.py
@@ -12,10 +12,13 @@ Office of Science.
 import glob
 import xarray as xr
 import numpy as np
+import pandas as pd
 import urllib
 import json
 from enum import Flag, auto
-from cftime._cftime import DatetimeGregorian
+from cftime._cftime import (DatetimeGregorian, DatetimeProlepticGregorian,
+                            DatetimeJulian)
+cftime_classes = (DatetimeGregorian, DatetimeProlepticGregorian, DatetimeJulian)
 
 
 class ARMStandardsFlag(Flag):
@@ -47,7 +50,7 @@ class ARMStandardsFlag(Flag):
 
 
 def read_netcdf(filenames, concat_dim='time', return_None=False,
-                combine='by_coords', use_cftime=True, cftime_to_datetime64=True,
+                combine='by_coords', use_cftime=True, cftime_to_datetime64=False,
                 **kwargs):
     """
     Returns `xarray.Dataset` with stored data and metadata from a user-defined
@@ -105,6 +108,8 @@ def read_netcdf(filenames, concat_dim='time', return_None=False,
     kwargs['concat_dim'] = concat_dim
     kwargs['use_cftime'] = use_cftime
 
+
+
     # Create an exception tuple to use with try statements. Doing it this way
     # so we can add the FileNotFoundError if requested. Can add more error
     # handling in the future.
@@ -141,7 +146,14 @@ def read_netcdf(filenames, concat_dim='time', return_None=False,
     # numpy datetime64.
     if 'time' in arm_ds.dims:
         desired_time_precision = 'datetime64[us]'
-        if (cftime_to_datetime64 and isinstance(arm_ds['time'].values[0], DatetimeGregorian)):
+        if cftime_to_datetime64 and isinstance(arm_ds['time'].values[0], cftime_classes):
+#            time = arm_ds['time'].to_index()
+#            arm_ds['time'].values = time.astype(desired_time_precision)
+
+#            print(type(arm_ds['time'].data))
+#            print(arm_ds['time'].data)
+#            time = time.astype(desired_time_precision)
+#            arm_ds.update(
             arm_ds['time'].values = arm_ds['time'].values.astype(desired_time_precision)
             try:
                 arm_ds['base_time'].values = \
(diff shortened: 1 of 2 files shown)


kenkehoe:
I'll investigate

kenkehoe:
I am having lots of troubles. Seems like my method of converting from cftime to numpy datetime64 is not preserving some cftime index. Anyone know how to correctly convert the times to numpy?

kenkehoe:
I'm having some issues with my pytest so I'm using this as a way to test. Let's not merge if this passes yet. I think we need to discuss this  a bit more.

kenkehoe:
After looking into this some more one of the issues this is fixing is not related to combine='by_coords'. I have found a solution but the bigger question is do we want to implement this for everything.

Using the decode_times keyword will look for a dimension that has a correctly formatted units string and magically convert to numpy datetime64. Our initial testing suggested that this worked correctly. But I don't think it is working correctly. Most ARM files have something like:

```
double time(time) ;
                time:long_name = "Time offset from midnight" ;
                time:units = "seconds since 2019-12-16 00:00:00 0:00" ;
                time:standard_name = "time" ;
```

This should set the time by reading the units string. We were assuming this was working because the times were correct after being read. But turns out that the units string in ARM files have timezone offset set to 0:00. This is following the correct standard so the files are OK. The issue is that if a timezone offset is in the units string the time part is ignored. So in this example the 00:00:00 0:00 is actually not used. But since we write files since that time it all just works. The issue is when there are files that using something like this:

```
double time(time) ;
                time:long_name = "Time offset from start of scan" ;
                time:units = "seconds since 2019-12-16 01:02:34 0:00" ;
                time:standard_name = "time" ;
```
In this case the time in Xarray dataset will start at 00:00:00. But if the units string is "seconds since 2019-12-16 01:02:34" then it all works correctly. I'm guessing this is a bug with Pandas? Not really sure.

The way to fix this is to not use decode_times keyword and manually read and convert time, but that means we can't use open_mfdataset. It would need to be open_dataset and each file will need to be converted and appended.

I found another way to fix this using use_cftime=True. This will use the CFtime library to read and convert the units string. It will work with timezone offset. But it will store values in CFtime in the Dataset. It took some digging to see how to correctly convert the times to datetime64 but I think I have that figured out.

The main question is what do we want to set as defaults for reading?

act/io/armfiles.py
```diff
@@ -192,20 +265,15 @@ def create_obj_from_arm_dod(proc, set_dims, version='', fill_value=-9999.,
         Depending on how the object is set up, sometimes the scalar values
         are dimensioned to the main dimension.  i.e. a lat/lon is set to have
         a dimension of time.  This is a way to set it up similarly.
-
```
zssherman: I believe this spaces are needed for sphinx to recognize.

zssherman:
Failure is doc related, the spaces removed in the pull request between section titles seems to be the issue, think spaces have to be added back in.

AdamTheisen:
@zssherman @rcjackson This PR is related to working around a potential issue with xarray and Ken and I have been discussing if it should be implemented in ACT or if we should wait for a fix from xarray.  There is potentially a simpler workaround for the time being that could be implemented I think.  I passed along to Ken and he was going to look into it as well as submitting an issue on xarray github...

kenkehoe:
I'm running my test code here to see if it matches what I get on my laptop. Looks like some of the tests may need to be updated because previous times don't match up anymore? Not really sure. Or there may be a rounding difference with the conversion from cftime to datetime64?

AdamTheisen:
This is definitely changing the values of some of the retrievals/corrections.  I'll clone it and see what might be going on with these

act/io/armfiles.py
AdamTheisen: @kenkehoe I just noticed where Zach's comment was.  In the docstrings, you removed the space between some of the sections.  Can you check and add those spaces back in as appropriate?

AdamTheisen:
@kenkehoe it looks like this method is changing how xarray resamples times...  for the cog/sog test, it is only when it gets here: do the data values change.

AdamTheisen:
It looks like this method is causing the times to offset slightly...

New: '2018-02-01T08:44:00.000016000'

Old: '2018-02-01T08:44:00.000000000'

I think this is causing differences in the resampling.

kenkehoe:
Just saw your comment about doc string spaces. I'll try to fix that one too.

AdamTheisen:
Thanks @kenkehoe !  This will be a big help, now knowing that xarray does not handle ARM time units correctly by default at the moment.

PR #162, opened 2020-01-02
ENH: New option for TimeSeriesDisplay to plot 2D data as 1D

AdamTheisen:
Also some documentation updates


AdamTheisen:
Figure_1

2D Stamp data plotted as line plots

AdamTheisen:
Expanding this pull request as there are some features added at the same time.  Also including 

- secondary y axis plotting ability with TimeSeriesDisplay
- fill between plotting capability addition (including secondary y)
- Added tests for fill_between and 2d to 1d capability

AdamTheisen:
Figure_1
Fill between option added to secondary y axis

AdamTheisen:
Requesting that Bobby take a look and also will ping @kenkehoe to check it out as well

rcjackson:
This is going to be a really great addition to ACT! I just have one suggestion.

Does the user have an option to prescribe legend labels for each line that comes out of the plot? I.e. right now I have no perspective as to which color corresponds to which row in the second dimension, but if the user gave a list of legend labels then this would be a lot easier to discern. In addition, we could do some default legend labels (i.e. 'row 0', 'row 1', ..., 'row n') so that the user can tell what ACT is plotting out by default. This should not be too difficult to implement.

AdamTheisen:
Good point @rcjackson.  I'll see what I can do.

kenkehoe:
Yeah I agree labels via legend is a good idea. What is the gray shaded area on second plot?

I've looked into second axis in the past and it's tricky with legend. Do we have a work around for that in this update?

kenkehoe:
I like this addition. Do we want to add code to convert a 2d data into multiple 1d data. dqlib has that code but it is not part of ACT. I can add that if we think it's useful. It's mostly xarray calls so we could also just have a place to document how to do that with xarray calls.

AdamTheisen:
@kenkehoe grey area is using the new fill_between code to add accumulated precipitation to the second y-axis.  

Right now the second y-axis option does not work with the regular legend call.  It is just a simple addition at the moment.  I can look into that more as well.

As for the 2d to 1D, it could be useful.  Right now, I am just passing in the 2D array to the pyplot.plot method and it is plotting it up fine.

AdamTheisen:
Pushing a commit for the legend functionality.

As for the secondary y-axis.  I think we would have to potentially add another value to our subplot_index.  Thoughts?

AdamTheisen:
SM_plot
Plot with Legend

AdamTheisen:
@rcjackson @kenkehoe what are you thoughts on proceeding with this PR as is and then I'll open an issue to see if we can determine a better method for the secondary y-axis???

kenkehoe:
I think going ahead with this PR and solving the secondary axis in a different PR is best.

Issue #164, opened 2020-01-03
u/v wind barb adding all colorbars to last subplot


cgodine:
When plotting multiple subplots using act.plotting.TimeSeriesDisplay.plot_barbs_from_u_v and adding a colormap to the subplots, all of the colorbars end up being plotted on the last subplot rather than their respective subplots:

![](https://dev.arm.gov/~godine/PLOTS/sgp/sgp915rwpwindcon/20200102/sgp915rwpwindconI8.a1.rwp_windcon.20200102.png)

I didn't notice anything in the code that I thought would be causing this. Any ideas?

rcjackson:
Chances are the plt.colorbar command is pointing to the wrong axis in the code. I'd check the ax= keyword in the plt.colorbar command.


________________________________

kenkehoe:
@cgodine did you check if ax= is the issue. If not I'll need to look into it.

cgodine:
@kenkehoe looks like it was indeed pointing to the wrong axis. Made a PR and can close this issue.

PR #173, opened 2020-01-08
FIX: Fixing how colorbars are plotted when there are multiple wind ba…

cgodine:
…rb subplots.

Referred to this in #164 and making a fix to point to the correct axes. 

New image:
image


AdamTheisen:
Code updates look good.  I'll open an issue as it does not look like we have a test for the plot_barbs_from_u_v function.

PR #175, opened 2020-01-10
Fixing issue when plotting data with lat and lon as vectors with day_ight_background set.

kenkehoe:
There is a problem when plotting data that have lat/lon as vectors when making the day night background. Just changed the method for getting data from .data to .values to return numpy array instead of dask object.


AdamTheisen:
Good catch on this.  Passing tests.  Merging...

PR #176, opened 2020-01-13
ADD: Options to customize ydata and colorbar labels

cgodine:
In certain datastreams, the height dimension is set to an integer-based index (ie, number of vertical levels of data rather than height in km) with height as a separate variable not used as a dimension. Using TimeSeriesDisplay.plot would result in the following:

I added an option to the plotting methods that don't explicitly set the y-axis to allow for a user to set height as the y variable provided the passed variable has the same shape. If it does not then it will default back to the original. 

I also added the option of setting the colorbar label manually. This already existed in time_height_scatter so I added it to the other plotting methods that called Display.add_colorbar. Here is an updated plot using these changes:

### DIFF
--- a/act/plotting/TimeSeriesDisplay.py
+++ b/act/plotting/TimeSeriesDisplay.py
@@ -327,6 +327,8 @@ class TimeSeriesDisplay(Display):
         else:
             ytitle = field
 
+        if cbar_label is None:
+            cbar_default = ytitle
         if len(dim) > 1:
             if use_var_for_y is None:
                 ydata = self._arm[dsname][dim[1]]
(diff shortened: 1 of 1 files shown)


kenkehoe:
Hey Corey. I just noticed that the colorbar title is not correct in my plots. Looking at the code I think we need to fix this. Can you take a look and see if this error is in your code. I don't think you introduced it but since you are making changes you can fix it with this PR. Looks like the Create labels if 2d as 1d code update may have caused this issue.

sgpmplpolfsC1 b1 mplpolfs_cross_pol 20200109

cgodine:
I think I found where the issue was. The default colorbar label was coming from the new y data rather than what was being plotted. Made a change to the default label:

AdamTheisen:
Thanks for catching this @cgodine .  Testing looks good, will merge.

kenkehoe:
Thanks Corey. I didn't have the brain power to fix last night in SFO airport. This should finish MPL!

PR #177, opened 2020-01-22
MNT: removed plot_quartiles since it was not being used

jhemedin:
Removed plot_quartiles from plot_stairstep_graph and plot_heatmap since it was not being used anywhere in the code.

It was listed as a variable in the definitions but not listed in the parameters or used anywhere in the code.


rcjackson:
This is a minor change and plot_quartiles isn't even in the docstring...so I'm going to accept.

PR #180, opened 2020-01-30
ENH: Napoleon

zssherman:
A nice version of numpy docs, removes the need for numpy doc extension as napoleon is built in with sphinx. This will fix the colon issue, and with the docs, I will add more commits to this pull request for the colon spacing. Overall cleaner docs. Down the road can invest in autosummary templates for modules, classes, and base so there would no longer be a need for rsts for all modules and submodules, just an index.rst

### DIFF
--- a/act/corrections/ceil.py
+++ b/act/corrections/ceil.py
@@ -1,3 +1,5 @@
+""" Functions for correcting ceilometer data. """
+
 import numpy as np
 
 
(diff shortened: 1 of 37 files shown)


AdamTheisen:
Thanks for all this work @zssherman!  Pinging @kenkehoe as well so that he's aware of the document updates.

kenkehoe:
Awesome thanks!

PR #184, opened 2020-02-06
ARM DQR Capability 

AdamTheisen:
Adding new function/program for working with ARM DQRs.  Also adding update for cropscape when the server is not functioning


AdamTheisen:
@rcjackson @zssherman I think I have the other web service issues handled in with this PR as well.  Let me know your thoughts on all this.

PR #185, opened 2020-02-10
FIX: Adding floor to barb step to avoid zero value

cgodine:
In instances where the number of x or y data points is less than half of the given num_barbs_x/num_barbs_y number, the barb_step_x/barb_step_y value will round down to zero and a ValueError will be raised when attempting to plot. Adding slice step floor value of 1 to mitigate this.


AdamTheisen:
@cgodine good catch!  This looks to be a minimal update to resolve the issue.

PR #186, opened 2020-02-10
MPL Correction

kenkehoe:
Updates to fix issues with reading data that is 1D in multiple files and returning as 2D. Also updated some processing to use broadcasting.

There is an issue with the updated reading routine where variables in the file as 1D are made 2D because of a change between files. Xarray is now doing this because of the keyword change we made. Also, updated a few attributes in the returned object.

Made variable names keywords to allow updates outside of changing code.

### DIFF
--- a/act/corrections/mpl.py
+++ b/act/corrections/mpl.py
@@ -12,7 +12,8 @@ def correct_mpl(obj, co_pol_var_name='signal_return_co_pol',
                 overlap_corr_var_name='overlap_correction',
                 overlap_corr_heights_var_name='overlap_correction_heights',
                 range_bins_var_name='range_bins',
-                height_var_name='height'):
+                height_var_name='height',
+                ratio_var_name='cross_co_ratio'):
     """
     This procedure corrects MPL data:
     1.) Throw out data before laser firing (heights < 0).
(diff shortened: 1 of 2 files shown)


act/corrections/mpl.py
```diff
@@ -58,64 +88,75 @@ def correct_mpl(obj):
...
         x_data[:, j] = x_data[:, j] * op[idx]
 
     # Create the co/cross ratio variable
-    ratio = (x_data / co_data) * 100.
-    obj['cross_co_ratio'] = obj[var_names[0]].copy(data=ratio)
+    if ratio_var_name is not None:
+        ratio = (x_data / co_data) * 100.
+        obj[ratio_var_name] = obj[co_pol_var_name].copy(data=ratio)
+        obj[ratio_var_name].attrs['long_name'] = 'Cross-poll / Co-poll ratio * 100'
+        obj[ratio_var_name].attrs['units'] = 'LDR'
```
AdamTheisen: The units should be '1' or 'unitless' right?

AdamTheisen:
I did some checks and it does not appear to change any of the cross/co ratio values.  I will add in a test for that in another PR

AdamTheisen:
Checks pass, merging in.

kenkehoe:
OK thanks.

PR #188, opened 2020-02-12
Fixing issue getting quality control variable names

kenkehoe:
Initially I tried to not use the qc_ at start of quality control variables since that is only an ARM standard, and ARM does not follow its own standards. But there is no reasonable way to code up all the minor differences with the attributes of ARM QC variables, so I'm adding in some code to also look for qc_ in the variable name.

### DIFF
--- a/act/qc/clean.py
+++ b/act/qc/clean.py
@@ -10,9 +10,18 @@ class CleanDataset(object):
         self._obj = xarray_obj
 
     @property
-    def matched_qc_variables(self):
+    def matched_qc_variables(self, check_arm_syntax=True):
         """
         Find variables that are QC variables and return list of names.
+
+        Parameters
+        ----------
+        check_arm_syntax : boolean
+            ARM ueses a standard of starting all quality control variables
+            with "qc_". This is a more robust method of getting the quality
+            control variables before the standard_name attribute is added.
+            If this is true will first check using attributes and will then
+            check if variable starts with "qc_".
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


kenkehoe:
Last time we had an error like this was because I moved some files around. I didn't do that this time. Do you have any further information on this error. I assume it's part of the HTML documentation generation?

AdamTheisen:
@kenkehoe  did you do a fetch on the latest ACT updates?

AdamTheisen:
@zssherman Pulling you in to see if you can see a fix for this that I did not

kenkehoe:
I did pull in the latest version from ACT master before making the new branch. But Git and I have a strained relationship so who knows what actually happened.

AdamTheisen:
Pulling in @rcjackson  as well to see if he can figure out what's going on here.

act/qc/clean.py
```diff
@@ -10,10 +19,19 @@ def __init__(self, xarray_obj):
...
-    def matched_qc_variables(self):
+    def matched_qc_variables(self, check_arm_syntax=True):
         """
         Find variables that are QC variables and return list of names.
 
+        Parameters
+        ----------
+        check_arm_syntax : boolean
+            ARM ueses a standard of starting all quality control variables
+            with "qc_". This is a more robust method of getting the quality
```
rcjackson: The underscore is your issue here. In RST, an underscore is a hyperlink. This needs to be "qc\_".

AdamTheisen:
Thanks @rcjackson !  @kenkehoe I pushed the fix.  Will merge.

kenkehoe:
Hmmm OK. Good catch. I'll try to remember this for the future.

PR #189, opened 2020-02-14
Fixing 2D heights

kenkehoe:
Adding correction if height dimension is 2D instead of expected 1D. This happens when xarray combines multiple files and the coordinate dimension changes values or shape between files. This is causing issues to subset height to positive values only.


AdamTheisen:
@kenkehoe Would you want to add a short sentence or two into the function description to note this behavior and that it will calculate the median of the heights in the reduction?

kenkehoe:
Yes I will add that documentation.

AdamTheisen:
Thanks @kenkehoe

PR #191, opened 2020-02-18
skewt plot issue

kenkehoe:
Fixing issue with calculating u and v wind speeds from wind speed and wind direction. Also updated to allow nans in Temperature.


AdamTheisen:
Looks like good updates to the function.  Will merge

PR #193, opened 2020-02-19
DOC: Updated cheatsheet to have ARM-DOE instead of ANL-DIGR

zssherman:


AdamTheisen:
Thanks for the quick update @zssherman

Issue #194, opened 2020-02-20
Missing dependency : appdirs


IreneCrisologo:
Hi, i just installed act from conda-forge and this error came up on import:

```
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
~/.conda/envs/cmac_env/lib/python3.6/site-packages/metpy/_version.py in get_version()
     13     try:
---> 14         from setuptools_scm import get_version
     15         return get_version(root='..', relative_to=__file__,

ModuleNotFoundError: No module named 'setuptools_scm'

During handling of the above exception, another exception occurred:

DistributionNotFound                      Traceback (most recent call last)
<ipython-input-11-d20e7d291629> in <module>
     31 import matplotlib.ticker as mt
     32 import matplotlib.font_manager as fm
---> 33 import act
     34 get_ipython().run_line_magic('matplotlib', 'inline')

~/.conda/envs/cmac_env/lib/python3.6/site-packages/act/__init__.py in <module>
      1 from . import io
----> 2 from . import plotting
      3 from . import corrections
      4 from . import utils
      5 from . import tests

~/.conda/envs/cmac_env/lib/python3.6/site-packages/act/plotting/__init__.py in <module>
     19 from .ContourDisplay import ContourDisplay
     20 from .WindRoseDisplay import WindRoseDisplay
---> 21 from .SkewTDisplay import SkewTDisplay
     22 from .XSectionDisplay import XSectionDisplay
     23 from .GeoDisplay import GeographicPlotDisplay

~/.conda/envs/cmac_env/lib/python3.6/site-packages/act/plotting/SkewTDisplay.py in <module>
     12 
     13 try:
---> 14     import metpy.calc as mpcalc
     15     METPY_AVAILABLE = True
     16 except ImportError:

~/.conda/envs/cmac_env/lib/python3.6/site-packages/metpy/__init__.py in <module>
     34 from ._version import get_version  # noqa: E402
     35 from .xarray import *  # noqa: F401, F403
---> 36 __version__ = get_version()
     37 del get_version

~/.conda/envs/cmac_env/lib/python3.6/site-packages/metpy/_version.py in get_version()
     17     except (ImportError, LookupError):
     18         from pkg_resources import get_distribution
---> 19         return get_distribution(__package__).version

~/.conda/envs/cmac_env/lib/python3.6/site-packages/pkg_resources/__init__.py in get_distribution(dist)
    480         dist = Requirement.parse(dist)
[...]

PR #195, opened 2020-02-20
ENH: Improved detection of properly installed MetPy

rcjackson:
This addresses #194, where I add in a few more exceptions in case there is an improper MetPy installation.


AdamTheisen:
Looks good!

PR #197, opened 2020-02-20
Error catching in day/night plotting

AdamTheisen:
Error catching in day/night background if a valid lat/lon is not included in the file


AdamTheisen:
Everything looks to be passing, minor error checking will merge.

PR #198, opened 2020-02-25
Sea Surface Temps from IRT

AdamTheisen:
There is a fair amount of updates in this one.  Ken K had passed along some code for converting surface and sky IRT values to sea surface temperatures.  These codes had a lot of overlap with the aeri2irt functions.  I've updated the existing functions to better adapt to multiple needs.  Overall updates include

- radiance_temperature_conversion was move from retrievals to utils under the name radiance_utils.  Function was also updated to work from different units
- The Planck_Converter was also found to have different constant values between scripts and was updated to follow that defined from a NOAA tool (documented in code)
- irt response function from aeri2irt was split out into a separate irt.py program
- Calculation of sea surface temperatures from IRT has been added to irt.py using dask for max efficiencyy
- tests have been added for sst calculation
- tests for aeri2irt have been updated due to the changes in values from the constant differences (0.01 K for the most part)

Note, a spot check of the output from this function was compared to output from Ken's scripts and found to be close in value (constant differences causing differences)


AdamTheisen:
pinging @kenkehoe  on this as well to see if there's any feedback

kenkehoe:
All sounds good to me. If something breaks I'll fix it.

rcjackson:
It looks ok to me and tests pass.

AdamTheisen:
Will merge in then. Thanks!

PR #201, opened 2020-02-26
ENH: Adding catch for nan data

AdamTheisen:
Noticed that the function fails on nan ir temperatures.  Fix for that


AdamTheisen:
Minor update.  merging

PR #202, opened 2020-02-28
Clean try

kenkehoe:
There is an issue when checking for variables incorrectly upconverted from int to float when xarray finds missing_value attributes. This function will try to revert the float back to int for things like QC and state type variables. This error was incorrectly catching errors while trying to guess the correct numpy dtype.


AdamTheisen:
All tests pass and looks like a good fix, merging.

PR #205, opened 2020-03-09
ADD: Ability to import Iowa Mesonet ASOS observations into xarray objects.

rcjackson:
I have added the ability to import ASOS data using the Iowa Mesonet Archive into an xarray dataset. Now we can specify regions or stations to download and get the data with ease.

### DIFF
--- a/act/discovery/__init__.py
+++ b/act/discovery/__init__.py
@@ -13,6 +13,7 @@ ARM Data Discovery.
 
     download_data
     croptype
+    get_asos
 """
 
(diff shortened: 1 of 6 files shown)


AdamTheisen:
Thanks @rcjackson!!  It looks like this adds some new requirements (six, urllib, urllib2).  I guess the urllibs are standard python right, so would not be an problem.  Would you want to update the requirements for six?

rcjackson:
I've went ahead and added six to the requirements, so it all should be good from there.

PR #208, opened 2020-03-19
ENH: ARM DQR Webservice Function

AdamTheisen:
Found a slight issue with the code at times when a qc variable was already in the obj.  This will check if a qc variable exists, if not, then add it.


AdamTheisen:
Very minor updates, merging.

Issue #37, opened 2019-04-19
day_night_background throws error when no location info in file


rcjackson:
We need to check if there is location data in the file before calling day_night_background. Right now it crashes if plotting the background with no location data, but we can just make it not do the background if there is no location information.

kenkehoe:
Additional feature request for this. 
- Handle vectors if it currently can't. (1D a 2D?)
- Handle vectors with the first value set to NaN. Need to look for first finite value and use.
- Handle a value for lat but not lon and exit gracefully.

kenkehoe:
I believe I put in a try to solve the initial issue for this Issue. I think it will handle scalar and 1D location data. Don't know about the rest of the bullets.

kenkehoe:
I will take a look at finishing this. You can assign it to me.

AdamTheisen:
Resolved with #211

PR #211, opened 2020-04-09
Fixing day_night_background when some values in vector are nan.

kenkehoe:
Finishing up issue #37 with location values and day night background plotting. This fix will handle vectors that contain nans. Also changed from throwing an error to showing a warning and returning without adding day/night background.


AdamTheisen:
Thanks @kenkehoe !  Tests pass and look to be a good fix

PR #213, opened 2020-04-13
Fixing issue with making numpy mask

kenkehoe:
In get_qc_test_mask() the returned value is expected to be a numpy masked array. Turns out the default behavior of numpy ma is to not make a ma if all mask values are to be set to False. This is not what we want. Need to set a keyword to force creation of ma even when all mask values are False.

Also updated the returned tuple from setting tests to have the data variable name as one of the members of the tuple. While in there fixed some documentation.

### DIFF
--- a/act/qc/clean.py
+++ b/act/qc/clean.py
@@ -476,8 +476,12 @@ class CleanDataset(object):
             return
 
         made_change = False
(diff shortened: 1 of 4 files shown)


AdamTheisen:
Passes checks and looks to be minor updates.

PR #214, opened 2020-04-22
Updated to have better default description for comparison test 

kenkehoe:
Updated to have better default description for test when the two variables to be compared are in the same xarray dataset.


AdamTheisen:
Minor updates and passes tests, merging.  Thanks @kenkehoe

PR #215, opened 2020-05-15
Adding delta test to QC

kenkehoe:
Updated other code to be more better including ensure the attributes set for limits are correct type. Added more tests to test_qc.py to ensure attributes are correctly set. I think I fixed a possible issue with missing value test and higher dimention data.

### DIFF
--- a/act/qc/__init__.py
+++ b/act/qc/__init__.py
@@ -17,6 +17,7 @@ and for applying tests to data
     qcfilter.parse_bit
     qcfilter.set_bit
     qcfilter.unset_bit
+    qcfilter.QCFilter.add_delta_test
     qcfilter.QCFilter.add_difference_test
     qcfilter.QCFilter.add_equal_to_test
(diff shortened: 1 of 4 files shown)


kenkehoe:
I don't understand the logs. Is it a failure with the test or flake?

AdamTheisen:
That's strange...  It's showing an error from the aeri code.

+eval 'flake8 --max-line-length=115 --ignore=F401,E402,W504,W605 '
++flake8 --max-line-length=115 --ignore=F401,E402,W504,W605
./act/retrievals/aeri.py:72:15: F541 f-string is missing placeholders

AdamTheisen:
I'll have to look into this a little more

AdamTheisen:
Changing a print statement that I think was causing the issue to raise an error instead.  We'll see if that works.

AdamTheisen:
Looks to be passing unit tests and looks like some good additions.  Will merge in.

PR #216, opened 2020-05-22
QC indexing

kenkehoe:
There is a strange issue getting indexes for setting tests from masked arrays. Switched to using numpy to get indexes and updated to allow boolean arrays in addition to index arrays for setting/adding tests. Should be a bit faster too.


AdamTheisen:
Thanks @kenkehoe !  For some reason I never got an email alert on this one.  I'll go ahead and merge and see if it got screened out of my email for some reason.

Issue #206, opened 2020-03-18
QC Barh Plot for 2D data


AdamTheisen:
When a user tries to plot 2D-QC data, it currently runs into an error.  It would be good to figure out plots similar to this.

![multi-1d_vars](https://user-images.githubusercontent.com/13421074/76995941-9d775400-691e-11ea-99e1-3afca744225b.png)

AdamTheisen:
@kenkehoe Pulling you in on this to see if you've thought about it.  Plot above is DQ Inspector

kenkehoe:
OK I had not thought about this. I am not surprised it causes an error. Do we want the same plotting response as DQ Inspector?

kenkehoe:
I've got this about 90% complete and just need to get the last bit updated. Should have a PR soon.

AdamTheisen:
@kenkehoe that is great!  Thanks for working on it!

kenkehoe:
I've created a PR for this update. Plot should be as good if not better than DQ Inspector.

![Figure_1](https://user-images.githubusercontent.com/8261054/83085396-07ac2380-a049-11ea-92e9-180d8d28490c.png)

kenkehoe:
This has been resolved. I don't see a way for me to close this issue.

AdamTheisen:
Thanks @kenkehoe !

PR #219, opened 2020-05-28
Qc flag block plot

kenkehoe:
Adding ability to plot QC for 2D variable that actually contains multiple 1D data. This should resolve issue #206 . Also fixed issue with y-axis label for force_line_plot option.


AdamTheisen:
Thanks @kenkehoe!  This looks to pass, would you be willing to put an example and/or a test together?

kenkehoe:
OK I've added an example script and test script.

AdamTheisen:
Thanks @kenkehoe for the test/examples!

PR #229, opened 2020-06-30
FIX: A fix for usage warning. I shouldn't have ran mpl2nc just should see if it exists.

zssherman:


zssherman:
@AdamTheisen This should fix both the mpl2nc usage message, but checking if mpl2nc and both pyart exist without using them.

AdamTheisen:
Perfect!  Thanks @zssherman

Issue #227, opened 2020-06-25
Day/Night Background Using Old Astral Function


AdamTheisen:
Currently the sun_utc function that is currently used in the day/night background creation is not available in the latest Astral releases (2.2).  We should work on updating to the latest functions if possible. @kenkehoe

kenkehoe:
Crap. OK I'll look into it. Thanks for heads up.

PR #230, opened 2020-07-01
Updating to use astral 2.2

cgodine:
Updating instances of astral to be compatible with the newest version of the library. Fix for #227


AdamTheisen:
@cgodine I had a feeling this was going to happen.  Conda only installs Astral up to 1.9.2 for some reason I am going to change our CI config files to pip install astral instead

AdamTheisen:
@cgodine This looks good now and is passing.  Will merge...  Thanks for updating!

PR #231, opened 2020-07-02
ADD: Function for calculating daytime/nighttime/dawn/dusk

AdamTheisen:
New feature to help with the calculation of daytime/nighttime periods for given locations.  Test and example have been included.  Tested across multiple locations but I'm sure there's something that may give problems in the future.


AdamTheisen:
All unit tests pass and tests across locations work.  Merging

PR #244, opened 2020-08-22
Shade time series

kenkehoe:
The previous attempt to overplot failing QC did not work as expected so I commented it out. The keywords where there but the code was commented out. I've reworked he code and it is much better. 

The basic plot with some failing QC will look something like
test_assessment_overplot

while a plot with two different variables plotted could be set up to look like this. I've given the option to have the overplot be plotted behind the data point so we can still see it. In this example I've done both behind and on top.
test_assessment_overplot_multi
I can't have it only add one Suspect to the plot legend since the second line is added after the first is plotted. But I think we should keep it this way because the user could choose to change the marker size, shape, color or location (behind or on top) and we need to correctly label that.

Lastly I've gotten it to work with the force_line_plot option as well.
Figure_1


AdamTheisen:
@kenkehoe thanks for this update!  Tests pass, code coverage has increased, and this fulfills a much needed capability.  Merging.

Issue #72, opened 2019-06-04
Filtering of data in plots (like a Py-ART gatefilter)


rcjackson:
Like what we have in the Py-ART GateFilter, we should have a way to filter out values when viewing/processing data. I see an ACT "gatefilter" in the future, but perhaps we can call it a DataFilter object.

kenkehoe:
I created the qcfilter.get_masked_data() method [located in qc.qcfilter file] to return filtered data based on QC status. Are you suggesting a filter to change the data in the object directly?

kenkehoe:
I think we should spend some time working on this. I suggest using the qcfilter.get_masked_data() to get a clean numpy array and then put that back into the object. This would then make plots with all the filtered data removed. That would be easier than not changing the data in the object and needing to call the method for every processing task. Since xarray has so many option for processing we should use them. The other question is can we convert the dask array containing number arrays to numpy masked arrays? Will that work with the xarray processing options?

kenkehoe:
This has been resolved with PR #245 .

PR #245, opened 2020-08-24
Data filter

kenkehoe:
Implementing the request under issue #72 . This will read the embedded QC variables and apply the results to the data in the xarray dataset. Has options to allow user to fix possible issues if they arise. Default is to set data to numpy masked array.


AdamTheisen:
Code looks good and tests pass, will merge.  Thanks for the addition!

PR #246, opened 2020-08-25
Coveralls update

kenkehoe:
Made more tests to get higher percentage for testing with coveralls. Actually found two issues so was worth the effort.


kenkehoe:
PEP8 stuff. Working on it.

AdamTheisen:
@kenkehoe looks like a good addition.  Only question I have is on the commented out code and what that was for.  Can we remove all that code that's commented out?

kenkehoe:
Oh good catch, that code should not be commented out.

AdamTheisen:
Thanks for the updates @kenkehoe , definitely helps with the code coverage

PR #251, opened 2020-09-03
Setting tests with flag values instead of flag masks

kenkehoe:
Found some issues when using flag values instead of flag masks.


AdamTheisen:
Thanks @kenkehoe !  Everything looks good and builds pass, merging in.

PR #254, opened 2020-09-15
Update to MPL Depol Ratio Calc

AdamTheisen:
Noticed an issue with the depol calc compared to what the mentor provided.  Updating that correction and the tests.


AdamTheisen:
@kenkehoe @ajsockol @cgodine Just want to make sure you are aware of this update.

kenkehoe:
How much of a difference is there with old code?

AdamTheisen:
@kenkehoe  it's really only with the ratio calculation.  In the old code it was (crosspol/copol) * 100.  This is crosspol/(copol + crosspol) * 100

kenkehoe:
OK I think this is fine.

PR #256, opened 2020-10-09
Bug fix for integer flag QC. Added tests for integer QC transfer to bitpacked QC.

kenkehoe:
Discovered that the code for updating from flag (integer) QC was not working correctly. Just updating the code to work correctly. Also updated the writer documentation.


AdamTheisen:
@kenkehoe all checks pass and this looks like a good update.  Will merge.  Thanks!

PR #259, opened 2020-10-21
Radiometer tests

kenkehoe:
The updated to FFT test messed with the 'time' coordinate attributes. Bug fix to make sure the attributes are not removed.


AdamTheisen:
Thanks for catching that @kenkehoe!

PR #260, opened 2020-10-21
Fixing issue when abs_limits set to a value of zero is interpreted wrong for applying values

kenkehoe:
While applying the abs_limits to a plot discovered a limit value of 0 is interpreted incorrectly as None/False so it is not bing used. Updated the code to correctly look for a value of None directly so a value of 0 can be used.


AdamTheisen:
All checks pass and this looks good.  Merging in.

PR #263, opened 2020-10-27
BUG: Trying one fix at a time

AdamTheisen:
Ship utils bug fix


AdamTheisen:
Merging this in since the coveralls seems to work.

PR #266, opened 2020-10-29
BUG:  Bug fix for when cartopy is installed but not importing correctly

AdamTheisen:


AdamTheisen:
Minimal bug fix, merging in.

PR #269, opened 2020-12-03
TST: Added tests and assert to raise coverage. Main focus stability indices in this PR.

zssherman:
Will do other tests in other PR to avoid overwhelming.

### DIFF
--- a/act/tests/test_correct.py
+++ b/act/tests/test_correct.py
@@ -18,13 +18,42 @@ def test_correct_ceil():
 def test_correct_mpl():
     # Make a fake ARM dataset to test with, just an array with 1e-7 for half
     # of it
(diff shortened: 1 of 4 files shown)


AdamTheisen:
@zssherman looks like just the following issue

./act/tests/test_utils.py:17:43: E712 comparison to False should be 'if cond is False:' or 'if not cond:'

AdamTheisen:
Fantastic work @zssherman !  Everything is passing so I will go ahead and merge

zssherman:
Hmm, I'm not sure how coveralls calculates its numbers, but coverage for stability indices went down from the PR, if I'm reading it right, overall coveralls went up 1%, i'll have to double check...

zssherman:
locally i was up to like 84%

PR #270, opened 2020-12-08
ENH: Updates for better unit test configuration, standardizing the di…

AdamTheisen:
…scovery tests and fixing some issues.


AdamTheisen:
@zssherman I updated the Astral check you had in.  Let me know what you think.  I also added a pytest config file to help with any future omissions we need.  I am removing the unit tests themselves and the _version.py file from the calculations.  Sound good?

AdamTheisen:
@jhemedin Just wanted to ping you on this since I couldn't request you as a reviewer for some reason.

act/tests/test_discovery.py
```diff
@@ -0,0 +1,63 @@
...
+    my_asoses = get_asos(time_window, lat_range=lat_window, lon_range=lon_window)
+    asos_keys = [x for x in my_asoses.keys()]
+    assert asos_keys == my_keys
+
+
+def test_get_armfile():
+    if not os.path.isdir((os.getcwd() + '/data/')):
+        os.makedirs((os.getcwd() + '/data/'))
+
+    uname = 'armlive_training'
```
AdamTheisen, line 48: @rcjackson @zssherman I was not too sure on including the test username and password in here.  Not sure the best/secure way to do this so that it runs on both travis and local machine.  Thoughts?

AdamTheisen:
@zssherman @rcjackson have you had a chance to take a look at this at all?  My main need for insight is the armlive credentials for testing and how we could better handle that.   Thoughts?

zssherman:
@AdamTheisen Sorry haven't had a chance to look. I'm not sure on the most secure way to handle it. Even though they are test username and password, I don't know if it would be fine to have them on github as part of the travis suite, @rcjackson your thoughts?

AdamTheisen:
Okay, @zssherman I think this is ready for review.  I had to update the stability indices tests to not be as precise as it seems like metpy calcs are changing with some updated versions

zssherman (approved):
@AdamTheisen Everything looks good to me, and the precision changes make sense.

PR #273, opened 2020-12-21
TST: Addition of contour tests

zssherman:
@AdamTheisen This is what I mentioned before with a parameter for contour, not sure if this is this route to go, however.

### DIFF
--- a/act/plotting/ContourDisplay.py
+++ b/act/plotting/ContourDisplay.py
@@ -24,7 +24,7 @@ class ContourDisplay(Display):
         super().__init__(obj, subplot_shape, ds_name, **kwargs)
 
     def create_contour(self, fields=None, time=None, function='cubic',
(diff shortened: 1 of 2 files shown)


AdamTheisen:
@zssherman  this looks good!  I would probably default it to do the filled contours though initially.  That's my only feedback.

Also, not sure what this error is from since you didn't do anything with the QC

FAILED act/tests/test_qc.py::test_arm_qc - IndexError: list index out of range

zssherman:
Yeah I wasn't too sure of that as well not sure why its failing. And sounds good, i'll change the default

AdamTheisen:
@zssherman it looks like the DQR webservice is down...  Checking on it now

AdamTheisen:
@zssherman everything is passing now and this looks good.  I will go ahead with the merge.  Thanks!

PR #278, opened 2021-01-22
Updating all baseline unit test images

AdamTheisen:
Changing travis ci pytest to include --mpl so that it properly runs the unit tests to compare images


zssherman (approved):
@AdamTheisen Looks good to me, tests still running though

PR #279, opened 2021-01-28
UPDATE: Added cartopy to environment download

jhemedin:
Added cartopy as a package to be downloaded when creating an environment using the environment.yml file. It's listed as a required dependency but was missing from the .yml file.


AdamTheisen:
Minor update, merging

PR #281, opened 2021-01-31
ENH: Bug fix to help with side by side plots on one row (1,2)

AdamTheisen:


AdamTheisen:
@kenkehoe @ajsockol @cgodine Do you have any side by side plots (multiple columns one row) that this would impact.  I was trying to create a side by side this weekend and it wasn't working as I would expect so made the update.  I.e. I would expect a subplot shape of (1,3) to have subplot indices of (0,0), (0,1), (0,2).  Thanks!

kenkehoe:
I don't think I have created any multi column plots.

ajsockol:
The only one that immediately comes to mind is the stamp_met_precip_comp plot for STAMP. Though, this plot is multiple columns and rows, not multiple columns and one row. If the bug didn't affect plots with multiple columns and rows, then I think the stamp plot should be okay.

AdamTheisen:
@ajsockol This should just affect plots with 1 row and multiple columns so I think that stamp plot should be fine.

zssherman (approved):
@AdamTheisen Looks good to me

Issue #274, opened 2021-01-08
Replace astral with a different solar calculation library supported on conda-forge


kenkehoe:
I was recently informed of a different Python library available for use to use for calculating solar positions and times called [pysolar](https://pysolar.readthedocs.io/en/latest/#location-calculation). It is still sub major release but a recent update in Oct. 2020 has me believing it is actively being developed. Initial impressions are that it may be more accurate with solar positions. The sunrise and sunset times are a little different than the Astral times. Comparison to [NREL solar calculator](https://www.esrl.noaa.gov/gmd/grad/solcalc/) are also slightly different. But I don't know how accurate the NREL solar calculator times are.

The documentation is not as good but it does seem to have a few benefits. The accuracy seems to be quite good. Speed for a single calculation is no different than Astral. It is on conda-forge but not the latest version, which is the one we would want. Currently highest version on conda-forge is 0.8. The latest version is 0.9 and available with pip install only.

One benefit is that it is more correct with dates of when the high latitudes transition to/from complete solar night/day. Astral currently does not handle this correctly and misses about 5 days. A second benefit is that it is capable of calculating TOA estimates of solar insolation and surface radiation values from a lat, lon , time entry. Also, able to make estimate of pressure and temperature changes due to elevation change. The additional calculations could be helpful.

If switching to pysolar gave us the ability to pull from conda-forge instead of pipy and more correctly calculates total solar night/day changes I think it is worth the effort to change. Right now I think we need some more digging.

AdamTheisen:
Discussed this on the ACT call today.  Initially, we are going to create a function in ACT that wraps around Astral to provide appropriate solar information.  This will allow greater flexibility for the future if we decide to switch solar libraries.  

In the mean-time, we are going to look at multiple options
- Continue looking at PySolar as an option
- Review skyfield library and if that could be an option
- Review the codes and see if it would make sense to code up the functions in ACT

kenkehoe:
I've started to work on this. There are some nice features to use with skyfield like it's vector processing. But it does not have the ability to produce a noon from a method call. I have it coded up to calculate noon as halfway between sunrise and sunset. This causes issues on full sun days as I can't calculate a noon value. My plan is to finish the refactoring to use a wrapper around the library with an option to use the library for the calculation. I got distracted by some other work so it will be a bit before I can make that PR.

brandon-rhodes:
[...]

PR #282, opened 2021-02-10
Phase 1 of Switch from Astral to Skyfield

kenkehoe:
I've refactored the code to use a new function to get solar sunrise, sunset, noon times instead of calling Astral directly. The new function uses skyfield library. The library uses vector processing so is much faster for some ACT operations. The next question is do we add additional libraries in the solar function to allow calling different libraries. I don't know if that is important right now. I did see both performance and accuracy updates with switching to skyfield. This is part of issue #274.


AdamTheisen:
@kenkehoe I'm seeing some interesting errors popping up.  Does skyfield need to download some data first or use a URL to update?

E           OSError: cannot download ftp://ssd.jpl.nasa.gov/pub/eph/planets/bsp/de421.bsp because <urlopen error ftp error: TimeoutError(110, 'Connection timed out')>
../../../miniconda3/envs/testenv/lib/python3.7/site-packages/skyfield/iokit.py:515: OSError

AdamTheisen:
@zssherman @jhemedin I might need your help on this one.  It looks like the skyfield library needs to download a file from an ftp site but ftp does not seem to be supported on Travis-Ci
https://blog.travis-ci.com/2018-07-23-the-tale-of-ftp-at-travis-ci

Have you ever encountered this before?  We might want to discuss on the skyfield issues board if they can make an option

zssherman:
@AdamTheisen @kenkehoe I have never encounter this before, but I wonder if there is a way or option in skyfield to download via https instead of ftp which will work with travis, its whether or not skyfield provides this.

zssherman:
I'm seeing that this possibly was fix in skyfield 1.31, which is not on condaforge, so many trying skyfield under pip in the ci environment ymls and see if that fixes the issue.

zssherman:
Wait nvm ignore my last point the recent skyfield should be on condaforge the outdated verzion buttons on the condaforge page threw me off.

zssherman:
@kenkehoe @AdamTheisen Can try for load('de421.bsp'), do load.open(https://ssd.jpl.nasa.gov/pub/eph/planets/bsp/de421.bsp') instead load.open(ftps://ssd.jpl.nasa.gov/pub/eph/planets/bsp/de421.bsp') There is an option to change the url, I believe open is an attribute to the loader.

zssherman:
I tried both neither, are working, still trying solutions

kenkehoe:
I've found a way to point to the file locally. I think putting the file in the repo will be best as that will not require users to download the file.

zssherman:
@kenkehoe Great! and for testing purposes and with travis stating they don't plan on addressing the ftp issue I think it is the best bet, the file is 12 MB so should be too big of an issue.

AdamTheisen:
@kenkehoe @zssherman Would it be best to try and get the up to date version from the FTP  and if that fails, then go to the local one?  Would it make sense to create a conf directory under the utils area or add this in the test/data area?

AdamTheisen:
Just so we have it, we can get the file through an https here: https://naif.jpl.nasa.gov/pub/naif/JUNO/kernels/spk/de421.bsp

AdamTheisen:
This looks good to me now.  I would like for us to work on fully removing astral before we do another release of the code.  I'll try and make that a priority over the next week

zssherman (approved):
Looks good to me @AdamTheisen @kenkehoe and I agree removing Astral before the next release would be a good idea.

kenkehoe:
@cgodine do you know if we are calling the older Astral functions? I didn't remove that old functions.

cgodine:
@kenkehoe i believe right now there are calls to each major version of astral (1.x and 2.x), where if 2.x is not installed it will use the older version method calls.

AdamTheisen:
@kenkehoe @cgodine I'll take a look at what's left to sort through and track in #274 .  Sound good?

kenkehoe:
@AdamTheisen OK. I think the only place Astral is still called is add_solar_variable() and the test scripts.

PR #284, opened 2021-02-16
Removing Astral Dependency Phase 2

AdamTheisen:
Updating add_solar_variable to use skyfield.  Also removing the references and dependency.  As part of this, the unit tests needed to be updated.


AdamTheisen:
I think this takes care of the astral dependency @kenkehoe

zssherman (approved):

Issue #253, opened 2020-09-10
Add method to update units in object


kenkehoe:
Adding a method to update the variable units in object. Could have option to look for all variable units in same category and make them all the same.

Could also have way to handle incorrect units that cause errors and use a look up table of known fixes.

AdamTheisen:
@kenkehoe Have you seen this at all? https://github.com/xarray-contrib/pint-xarray

kenkehoe:
@AdamTheisen thanks for the link. This looks interesting.

I think we will need to discuss this a bit as there are some promises for this to help, but also some issues.

Issues:
- It appears installing pint-xarray will force us to use xarray=>0.16. We are currently at v0.12 for some other issues. I have not gone back to test again so maybe we can move to v0.16.
- I think there is something still wrong with pint not liking the use of "1" for unitless variables or the symbol of "%". This is a known issue we have code to get around. This does cause issues with code that works on the full dataset so we will need to loop over each variable DataArray to catch the errors and continue even if we are only trying to change on DataArray by calling the method on the dataset.
- I don't like how pint will remove the units attribute when the DataArray values are changed to a pint object, and then add it at the end of the attribute list when it adds it back in. This is not a big deal but it does make an unexpected change.
- In our existing unit conversion function I had to be careful about pint converting data type from int to float and manually convert it back when we don't want that action. I assume that is still part of this and something else we need to be careful to manage.
- Pint will accept a unit of 'degF' but when it writes the units back into the DataArray it writes it as "degree_Fahrenheit" While correct it is not what I asked pint to do. I'd prefer it keeps the unit string I provided of 'degF'.

I think we should write a method to perform the looping on the full dataset by working on each DataArray to ensure the data type and units are set as we expect and catch the ensuing errors. But yeah we should be able to use this new library.

kenkehoe:
I looked into this a bit for a different reason. I would like to still create a method to update units for multiple variables in the object to make matching other data easier. I think that would be straight forward for a method to loop through and update units using existing function we have convert_units(). But I have run into an issue with updating coordinate variables. Using newer Xarray will produce a warning when using *.values to assign new values after unit transformation. It requires using .assign() or .assign_coords() which returns a new object. This could be problematic for a method. I'll look into this pint-xarray to see if it handles coordinate variables correctly. If so we may want to go this route.

PR #285, opened 2021-02-17
Units

kenkehoe:
This PR is to implement #253 . This will be a nicer way to change the units in the dataset than manually changing them. This will allow changing all variables with a loop that are compatible with the request unit or a select set of variables.

I'm having an issue with a different function test failing now that was fine before. Going to make this PR to see if TravisCI has the same failure.

### DIFF
--- a/act/plotting/SkewTDisplay.py
+++ b/act/plotting/SkewTDisplay.py
@@ -315,6 +315,8 @@ class SkewTDisplay(Display):
             u_red[i] = u[index].magnitude * getattr(units, u_units)
             v_red[i] = v[index].magnitude * getattr(units, v_units)
 
+        u_red = u_red.magnitude
+        v_red = v_red.magnitude
         self.SkewT[subplot_index].plot(p, T, 'r', **plot_kwargs)
         self.SkewT[subplot_index].plot(p, Td, 'g', **plot_kwargs)
         self.SkewT[subplot_index].plot_barbs(
--- a/act/tests/test_plotting.py
+++ b/act/tests/test_plotting.py
@@ -135,11 +135,10 @@ def test_skewt_plot():
 
 @pytest.mark.mpl_image_compare(tolerance=30)
 def test_skewt_plot_spd_dir():
-    sonde_ds = arm.read_netcdf(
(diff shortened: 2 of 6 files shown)


kenkehoe:
I don't have much for unit conversion examples. Do you think we should add that?

AdamTheisen:
@kenkehoe Examples would be good, but I don't think it is necessary as part of this PR.

@jhemedin @zssherman would either of you have time to take a look at this new class and give it a run through.

kenkehoe:
I've added an example since it's not obvious the method exists.

AdamTheisen:
@kenkehoe got a chance to look at this and the main question i have is does this need to be a Class?  Does that do anything for us @zssherman ??

kenkehoe:
I wanted to make this a method so the function does not need to be imported. We could put in a different class name, but I thought the name was generic enough to add more in the future.

zssherman:
@AdamTheisen I'm not too familiar yet with accessors, but it almost looks its a class so that when you read a file, the object can just called change units instead of calling change units and imputing the obj as a parameter and does the change inplace, but im not sure 100% sure. The function itself looks fine as does the tests and example. But seems other unit tools in ACT are functions, so I think this is more a consistency thing, I'm not sure on which route to take or if this has to be a class.

zssherman:
I commented right when Ken's message loaded for me haha

zssherman:
To simply the code and more on the side of Python standards, a class might be too much for what the code is trying to do, but for the simplicity on users, the class might be simpler and easier access, its more on which we prefer and might have to vote on.

zssherman:
Also if we plan on adding more functions within the scope of the class, then a class would be preferable.

AdamTheisen:
Thanks @zssherman.  @kenkehoe if you expect that we will have more functions to add under this class in the future, then I would say we can leave as is.  Will wait until tomorrow for final comments before merging

kenkehoe:
I think we will add additional utils methods in the future. They will be called under act.utils.* methods. I think we can call the class anything you want without breaking the code since the class name is not called directly.

AdamTheisen:
Given the expectation that we will add more under this in the future, I am fine with merging as is.

PR #289, opened 2021-02-19
Bug fixes to skewt plotting try 2

AdamTheisen:
I ran into errors when trying to do multi-panel skewt plots. These should fix it and also enable the figsize to work appropriately.  Adding a test for multi-panel skewt


AdamTheisen:
Going to go ahead and merge since it is passing tests

Issue #252, opened 2020-09-10
Add method to update object to CF standards


kenkehoe:
There are are few CF standards that are not part of ARM files or others. Can add suggested attributes to be CF compliant.

kenkehoe:
This can be assigned to me.

AdamTheisen:
@kenkehoe I believe your latest PR satisfies this issue right?  As you mentioned, there are likely some that were missed, but we can update as we find them.  THoughts?

kenkehoe:
Affirm, the last PR covered this issue.

PR #291, opened 2021-03-04
CF compliant

kenkehoe:
Making some updates to the netCDF writer to add some additional attributes needed to write a CF compliant data file. This works with issue #252 . I've added attributes for the things I know about, but I would assume I missed one or two. Also, discovered an issue with the QC cleaner that I have updated.

### DIFF
--- a/act/io/armfiles.py
+++ b/act/io/armfiles.py
@@ -351,9 +351,9 @@ class WriteDataset(object):
         self._obj = xarray_obj
 
     def write_netcdf(self, cleanup_global_atts=True, cleanup_qc_atts=True,
-                     join_char='__', make_copy=True,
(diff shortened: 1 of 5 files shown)


kenkehoe:
Soooo, I forgot to add testing to this. I'll do that next. But I think something strange is going on with the act/utils/conf/de421.bsp file used with skyfield. We are getting a lot of errors about it not getting closed.

AdamTheisen:
Thanks @kenkehoe I'll take a look at that

kenkehoe:
Looks like something got hung-up with the 3.7 testing and it timed out. I'm going to add some more test coverage and then make another push to run Travis again.

AdamTheisen:
Tests are passing and the code looks good.  Thanks @kenkehoe !

PR #293, opened 2021-03-05
Updates to QC Block Plot

AdamTheisen:
Update to get a color from matplotlib CMAP if the assessment is not defined.  Adding in assessment color for "Acceptable"


zssherman (approved):
@AdamTheisen Looks good to me!

PR #295, opened 2021-03-09
ENH: Updating to use latest matplotlib version

AdamTheisen:
Removing the restriction on matplotlib version.  Testing on local machines shows version 3.3.2 works well.


AdamTheisen:
Something is failing in the test_barb_sounding_plot with the latest matplotlib version.  Will need to investigate

AdamTheisen:
Just to note as I won't get to this today.  It looks like the matplotlib convert_xunits does not handle the time well that's passed in to pcolormesh from the ACT code.  We might need to revamp this function somehow.  Current testing shows that it's converting the MaskedArray to a list in this function and then checking the shape of the list using .shape....  Unsure if it's a bug in matplotlib or in how we are doing things.

zssherman:
Yeah I see what you mentioned, X is wrapped in a list as compared to Y, but it almost looks like its occuring before the pcolormesh code, just can't pinpoint whats causing it yet..
X = [array([13167.21041667, 13167.21041667, 13167.21041667, 13167.21041667,
       13167.21041667, 13167.21041667, 13167.2...42686632, 13168.42686632, 13168.42686632,
       13168.42686632, 13168.42686632, 13168.42686632, 13168.42686632]), ...]
Y = array([[   4.9000001,   57.5368422,  110.1736843,  162.8105264,
         215.4473685,  268.0842106,  320.7210527,  373...36.5421053,  689.1789474,  741.8157895,  794.4526316,
         847.0894737,  899.7263158,  952.3631579, 1005.       ]])

AdamTheisen:
@zssherman what I found is that on line 5571 in _axes.py that the code below actually converts the numpy masked array of datetimes into a list for some reason.  I tried digging into the convert_xunits which points to convert_units which points to something else but haven't actually found the criteria for conversion.  We use PCOLORMESH in our 2D time series plots so I may see if I can restructure this function to be similar to how we call pcolormesh there.  

`X = self.convert_xunits(X)`

zssherman:
Error was connection issue, I restarted the tests

AdamTheisen:
@zssherman this was due to the issue in #298  and the change to the API.  All should be good now with that merge.

zssherman (approved):

zssherman:
Looks good to me!

PR #299, opened 2021-03-12
Updating requirements to shift Cartopy to optional dependency like we initially intended.

AdamTheisen:


AdamTheisen:
#112 added cartopy to the requirements, but our code is now set up so that cartopy is an optional dependency.  Removing cartopy from the requirements document.

zssherman (approved):

PR #303, opened 2021-03-29
removing numpy alias for string data type

cgodine:
Starting with NumPy 1.20, the 'np.str' alias is deprecated. Switching to the built-in str type to fix the DeprecationWarning.


cgodine:
I believe the DQR web service went down while CI was running. @AdamTheisen do I need to commit again to get CI to re-run?

AdamTheisen:
@cgodine I just restarted it but will have to look at workarounds for that in the futrue.

PR #304, opened 2021-03-30
ENH: Updates to get_armfiles for when the API is not functioning.  Al…

AdamTheisen:
…so just added in for it to return filenames


jhemedin (approved):
All looks good and working.

PR #312, opened 2021-04-09
Fixing issue calculating solar noon, sunrise, sunset at high latitude.

kenkehoe:
The way these values are calculated it needs to find an instance. At high latitudes it needs to look a larger number of days forward and backward to find an instance when there are polar night/day. Increasing number of days to calculate to find matches.

### DIFF
--- a/act/tests/test_utils.py
+++ b/act/tests/test_utils.py
@@ -402,6 +402,11 @@ def test_get_sunrise_sunset_noon():
     assert sunset[0].replace(microsecond=0) == datetime(2018, 2, 1, 17, 24, 4)
     assert noon[0].replace(microsecond=0) == datetime(2018, 2, 1, 8, 2, 10)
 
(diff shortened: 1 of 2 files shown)


kenkehoe:
Hold off merging. Still looking into an error.

kenkehoe:
OK it's working correctly now. Not sure what happened with Travi-ci. Is there a way to start it again?

AdamTheisen:
@kenkehoe There's something going on with travis-ci and it's not letting me restart.  I'll give it a little and retry

AdamTheisen:
@kenkehoe do you have an example dataset for testing or could we use the NSA or MAR files in the test area?

kenkehoe:
I believe it is being tested, just not above 80 degrees north which is why there was a problem with MOSAiC data. I think it's good without testing that specific data. Otherwise would be OK to test with hardcoded values in the test script.

AdamTheisen:
I can work on creating a test for it with some fake data.  Shouldn't be too hard.

PR #317, opened 2021-04-19
ENH: Adding more tests for I/O module and fixing some bugs that came up

AdamTheisen:
Adding some tests and sample files for some of the harder to get at testing for the IO modules.  Adding some additional flexibility for file_date/time  if the file name is not ARM format


AdamTheisen:
The discovery module normally adds 0.6% after merges so this should get us to 90%

zssherman (approved):
Everything looks good to me!

PR #322, opened 2021-05-06
BUG: Bug fix for the add_dqr_to_qc.

AdamTheisen:
It was returning local times instead of utc times

### DIFF
--- a/act/qc/arm.py
+++ b/act/qc/arm.py
@@ -83,6 +83,7 @@ def add_dqr_to_qc(obj, variable=None, assessment='incorrect,suspect',
         url += ''.join(['&searchmetric=', assessment,
                         '&dqrfields=dqrid,starttime,endtime,metric,subject'])
 
+        print(url)
         # Call web service
         req = requests.get(url)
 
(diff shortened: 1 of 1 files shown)


AdamTheisen:
@zssherman Just wanted a second set of eye here to make sure this is correct.  The values I get correspond to the DQR this is for https://www.archive.arm.gov/ArchiveServices/DQRService?dqrid=D210405.5 so I believe this is now correct

fourth_vis

zssherman:
@AdamTheisen I'm not familiar with the DQR system, how do I compare values? Probably is obvious but just slightly confused

AdamTheisen:
Sorry @zssherman, should have explained that more.  In the DQR linked, there should be a list of times.  For this particular one, only one of those date/time ranges will correspond to 3/29/2021 which is when the plot is from.  

2021-03-29 13:00:00 | 2021-03-29 20:10:00 | Incorrect

The times on the plot should line up.  I also was not sure if you used these fromtimestamp functions or timezones much before and just wanted to make I wasn't making another major mistake!

zssherman:
Ah gotcha, I think I understand, so the red bar for shadowband misalignment? Which on the plot looks to match the DQR from 13:00 to 20:10, I believe it is good, unless i'm still misunderstanding.

AdamTheisen:
@zssherman Yeah, those should line up.  I could have sworn I verified this before in the past but it was off when I was doing some analysis for this case. I don't think there's a larger issue at play but just wanted a second set of eyes.

PR #325, opened 2021-05-12
ARM Filename Check to get date and time

kenkehoe:
This is to fix an issue with parsing the filename to get date and time when a sub-directory also has a ".". While I was in there changed to use re to be more sure about filename for date and time.


AdamTheisen:
@kenkehoe This error is popping up due to the new matplotlib version.  Taking a look

AdamTheisen:
Looks good to me, adding @zssherman as a reviewer for approval and merging.

zssherman (approved):
@AdamTheisen Everything looks good to me!

PR #327, opened 2021-05-20
NOAA data

kenkehoe:
Sorry about so many changes in one PR. I was working on two things at once and it is easier to pull them all in at once. This has updates to the NOAA data reader to fix some minor issues and allow reading multiple CSV files. Also, just discovered how to correctly add a new QC variable to a Dataset as a dask array instead of numpy array to match the data variable type.


kenkehoe:
I'm looking into the failed tests.

jhemedin (approved):
All looks good. Accepting and merging

PR #328, opened 2021-05-24
Updates for consistency across all modules

AdamTheisen:
PR to work towards a v1.0.0


AdamTheisen:
@zssherman should we merge this and make it a priority to resolve the documentation issues before releasing V1.0?  This is mainly updating codes and docstrings in each file to be consistent.  Also changing the name of the cropscape module to be more consistent with the other codes.  Also removing the release history since that's in Github and we have to manually update in ACT code.

zssherman:
@AdamTheisen Yeah I think that would be fine. I don't think fixing the rest of the documentation will be an issue.

PR #330, opened 2021-06-03
DOC: Overhaul of whole documentation suite.

zssherman:
So after much sweat and tears and anger haha, I have overhauled the whole documation suite for ACT. What this does is allow summary tables for all functions, classes and class functions. This also autogen mostly everything, so no more 'act.plotting.xsect' etc docstrings at the top of modules and __init__.py autodocs with do that for us. Makefile and make.bat had to be overhauled to work with the new gen system. rsts for each submodule have been removed, only an index.rst is needed now. _templates folder was added into source, with templates for base, class and modules. These are needed for autodocs to know how to autosummary everything in ACT. But that is everything for the doc side of things.

For this to work and to be consistent with standards anyways, one of the reason we were having import errors with some plotting submodules and doc gen was because the submodule is the same name as the class withing which breaks many different things. To fix this, the submodules should be undercase, the classes as they should be uppercase, so this is what I changed and fixed. I need to update the examples as well as the imports are now broken, we can discuss this, but this the best path forward, submodules shouldn't be uppercase.

I'll update the examples as well, @AdamTheisen or @jhemedin if one of you or both could double generate on your end and make sure this works.

Side note if anyone is using the plotting suite in a system, they will need to be made aware of the import plot changes.


zssherman:
Updated note: the plotting stuff should matter because were are __init__.py, so the imports shouldn't change at all I think, we'll see what the tests do.

zssherman:
Everyone who contributes to ACT as well should check the changes in __init__.py for every submodule as well as modules themselves to see what is only needed now for module docstrings. Only a background info is needed now so for example:

```
"""===============================act.retrievals (act.retrievals)===============================.. currentmodule:: act.retrievalsThis module contains various retrievals for datsets... autosummary::    :toctree: generated/    aeri2irt    calculate_dsh_from_dsdh_sdn    calculate_irradiance_stats    calculate_net_radiation    calculate_longwave_radiation    calculate_precipitable_water    calculate_stability_indicies    compute_winds_from_ppi    generic_sobel_cbh    sst_from_irt    sum_function_irt"""
```

AdamTheisen:
@zssherman  This is definitely an improvement and cleaned up.  A couple things I'm seeing.  In the sidebar navigation, it looks like the modules are being listed twice.
Screen Shot 2021-06-03 at 10 00 03 AM

In the plotting API pages, I'm not seeing any of the sub functions.  I.e for TimeSeriesDisplay, I'm not seeing a way to view the plot, wind bard, day night background ,etc pages. 

Thoughts?

zssherman:
I might have to check if I accidently messed the depth up, which could be causing that duplication, I'll see if I can find what's causing that. And timeseries as well, might have missed a generation in the index of something I'll check and see what's causing both.

zssherman:
I am not able to reproduce the doubling, maybe do a rm -r build and rm -r generated in docs and source/api if you haven't? I do have timeseries appearing as well with all functions. However i encountered a different error were modules with __init__ imports of:
from . import clean don't document, I will have to see if changing the imports will fix that.

zssherman:
Screenshot from 2021-06-03 12-48-48

zssherman:
The pythonic way of doing this is to no longer in __init__ do from .ceil import * or from . import ceil, but instead do:
from .ceil import ceil, foo, etc

Which adds to adding more to the __init__ but more info is better

zssherman:
But that fixes the last issue im encountering.

AdamTheisen:
@zssherman that is strange.  I cloned your branch and have been doing the build from that.  I am going to update all my packages and see if that resolves anything because I am definitely seeing than the act.qc listing you have above

zssherman:
@AdamTheisen I updated the __inits__. For the images, some of the examples are just print outputs which is why there are no thumbnails for some, others I still haven't figured out, I'll try a couple different things in a different PR.

AdamTheisen:
@zssherman can you check the IO documentation?  I wasn't seeing anything in the docs so updated the __init__ to 

`"""
This module contains procedures for reading and writing various ARM datasets.

"""

from .armfiles import  *
from .csvfiles import *
from .mpl import *
from .noaagml import *
`

and it ended up pulling in Path and datetime into the documentation.  Are you seeing this as well?
Screen Shot 2021-06-04 at 4 39 48 PM

zssherman:
@AdamTheisen oops yeah that's my bad I missed the io init and let me see if I can reproduce the imports if datetime and path. If so, I might just do a regular import of that submodule

zssherman:
@AdamTheisen Yeah, so not sure what was going on there, so i'm guessing when those imports path and datetime are called and used, they are being picked up by autosummary, the way around it I changed and I guess more pythonic anyways, is I specified each module and where they are coming from, which should fix the issue now.

AdamTheisen:
@zssherman Thanks for all the effort on this!  I think this is good to go but will await your final say on it.  I made some updates to the index.rst file to be consistent with the README

zssherman:
@AdamTheisen I can't think of anything else to change or add at the moment, I think its all good to me to merge!

PR #331, opened 2021-06-09
Data utils

kenkehoe:
Adding three new functions to correct temperature and pressure for differences in height.


AdamTheisen:
Thanks @kenkehoe !  Looks like the unit tests you added are failing

FAILED act/tests/test_utils.py::test_convert_to_potential_temp - assert False
FAILED act/tests/test_utils.py::test_height_adjusted_temperature - assert False

With the specific errors visible here: https://travis-ci.org/github/ARM-DOE/ACT/jobs/774063481#L3187

kenkehoe:
Well this is interesting. The tests pass on my laptop. I adjusted the atol=0.01 to see if that was the issue, but that didn't fix it. I'll keep looking into this.

AdamTheisen:
@kenkehoe The only other item i see is that you're doing a direct import of metpy but it is an optional dependency. We need to do something like the sonde.py retrievals import

https://github.com/ARM-DOE/ACT/blob/master/act/retrievals/sonde.py#L10

kenkehoe:
OK good point. I've updated to be an optional import.

AdamTheisen (approved):
I've made some updates to expand the cover coverage and fix one possible bug.

PR #332, opened 2021-06-09
Deprecation Warning fix

jhemedin:
Updated the TimeSeriesDisplay.plot() deprecation error when X,Y, and C have the same dimensions in pcolormesh(). 

Added a variable called set_shading that can set the shading parameter to auto and lets matplotlib check if shading should be set to 'flat' or 'nearest' depending on  dimensions.

Default is True but user can set it to false and use kwargs to set the shading parameter.


AdamTheisen:
@jhemedin Thanks for the fix on this!  Looks like just one PEP8 issue

./act/plotting/timeseriesdisplay.py:475:1: W293 blank line contains whitespace

zssherman (changes requested):
Code wise, everything looks good, only thing that needs fixing is the flake8 whitespace that travis flagged.

jhemedin:
removed whitespace, tests are rerunning

AdamTheisen:
@jhemedin that does not seem to have fixed the issue on 475.  Can you take another look?

./act/plotting/timeseriesdisplay.py:475:1: W293 blank line contains whitespace

PR #337, opened 2021-06-11
armfiles

kenkehoe:
Fixing issue with fixing time and making xarray happy with keyword selections. Turns out one keyword use now is causing a warning and will be an error in next xarray release. Also fixed issue when setting time values without useing .values

### DIFF
--- a/act/io/armfiles.py
+++ b/act/io/armfiles.py
@@ -17,9 +17,9 @@ from pathlib import Path
 import re
 
 
-def read_netcdf(filenames, concat_dim='time', return_None=False,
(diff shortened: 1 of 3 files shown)


AdamTheisen:
@kenkehoe There might be an issue with the build but travis-ci site has migrated and we will need to update our testing to point there as well.

AdamTheisen:
@kenkehoe we can take a look once we get things migrated over to travis-ci.com since it seems like it's unrelated to your updates and likely a dependency issue

AdamTheisen (approved):

PR #338, opened 2021-06-16
ENH: Updating readme

AdamTheisen:
Testing out the travis build to make sure it's not an issue with the PR being open while migrating.


rcjackson:
I had to activate the plan in Travis CI to use the free plan. That is what was stopping it from doing builds, so it should work now.

Issue #323, opened 2021-05-06
ARM DQR to QC adding multiple qc tests for same DQR


AdamTheisen:
When i tried plotting up multiple days of a dataset with DQR periods spread across those days each periods was added as a separate QC flag in the qc variable.  I.e. there were 5 tests with the same name with different corresponding times.  We need to look at a more robust way to add these DQRs to the qc variables

AdamTheisen:
![dqr_issue](https://user-images.githubusercontent.com/13421074/117345869-abfb3480-ae6c-11eb-8826-600e7c74a86c.png)

Plot showing issue

kenkehoe:
Ummm, that is strange. I'll take a look.

kenkehoe:
I believe this has been resolved.

PR #341, opened 2021-06-18
Fixing issues when adding DQRs to ACT object

kenkehoe:
This is to fix issue #323 . Turns out it was adding a new test for each time range.

While testing found issue overplotting failing data and using absolute limits.

Since I was in there, decided to reduce warnings by using finally: with plot tests to close plot after used and capture warning with calculation of u & v direction.

### DIFF
--- a/act/plotting/skewtdisplay.py
+++ b/act/plotting/skewtdisplay.py
@@ -211,8 +211,10 @@ class SkewTDisplay(Display):
         # Make temporary field called tempu, tempv
         spd = self._obj[dsname][spd_field].values
         dir = self._obj[dsname][dir_field].values
-        tempu = -np.sin(np.deg2rad(dir)) * spd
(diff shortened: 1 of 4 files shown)


AdamTheisen:
@zssherman are there any issues with wrapping the figure returns in the unit tests with a try?

kenkehoe:
I'm looking into a warning that was created by changes to previous PR. Maybe hold off on approval just yet.

zssherman:
@AdamTheisen I don't think so, It should be fine

kenkehoe:
OK, after looking into this I think read_netcdf() is correct and our production system is just behind on current updates in GitHub. If this is good to merge I say go ahead.

Can we also then make a conda-forge release for 1.0.3 to resolve the issues on our production system?

AdamTheisen:
@kenkehoe once @zssherman approves and merges in, I can work on a new micro release.

zssherman (approved):

PR #343, opened 2021-06-30
Numpy float 

kenkehoe:
Numpy has updated the way to set data type and there is currently a warning for one last place we use the old format. This fixes that warning. Also, fixing issue setting yrange on plots when there are inf values.


AdamTheisen (approved):

AdamTheisen:
Looks good, thanks @kenkehoe !

PR #344, opened 2021-07-01
Bug fixes for the pyart converter and other small updates

AdamTheisen:


AdamTheisen:
Got this working finally after mixing some branches up.  Going to merge.

PR #355, opened 2021-08-03
QC dask

kenkehoe:
This PR contains updates to the QC code to use Dask processing to try and speed up the initialization and setting of tests. It's not showing faster results on initial testing, but should be faster under some circumstances.

Also, added two additional tests using normal distribution assumption with a Grubbs style test. Requires an additional optional library to do the testing.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -69,6 +69,7 @@ Optional Dependencies
 * `MPL2NC <https://github.com/peterkuma/mpl2nc>`_ Reading binary MPL data.
 * `Cartopy <https://scitools.org.uk/cartopy/docs/latest/>`_  Mapping and geoplots
 * `MetPy <https://unidata.github.io/MetPy/latest/index.html>`_ >= V1.0 Skew-T plotting and some stabilities indices calculations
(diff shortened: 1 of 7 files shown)


kenkehoe:
Appears this failed on import of the new library for the two new tests. I added the new library under the README.rst but not in environment.yml file. I assume I need to add the new library in environment.yml ?

zssherman:
@kenkehoe Yeah, the new library needs to be added to the environment files in the CI directory, do we want to add this as an optional dependency or a dependency?

kenkehoe:
This should be an optional dependency. I've added it under the optional part in the README.rst file. I assume that is where we have the listing of required vs. optional dependencies?

zssherman:
@kenkehoe Yeah the README works, the module needs to be added to environment-3.7 and 3.8  as well under the continuous integration folder for travis to install, but everything else looks fine to me.

kenkehoe:
So do we want it listed in environment.yml file if it's an optional dependency? I'll add to the environment-3.7 and 3.8 files.

zssherman:
For the environment file, I don't think so, looks like the other optionals aren't there. Think that is just base dependency environment for users.

zssherman (approved):
Looks good to me @kenkehoe @AdamTheisen !

PR #358, opened 2021-09-15
ADD: Adding noaa psl wind profiler reader.

zssherman:
Code for reading the noaa psl wind profiler data.

Few things, this code is bound to change, as dealing with this data is tricky. I wanted to get the code going in a PR as I work on finishing the unit tests, so if we decide to change anything, I can adjust the unit tests accordingly.


zssherman:
Looks like the failing test is a skewt plotting fail due to metpy.

AdamTheisen:
@zssherman one last item on this it looks like

./act/io/noaapsl.py:37:31: E712 comparison to True should be 'if cond is True:' or 'if cond:'

Once that's fixed, I'll do some testing

zssherman:
@AdamTheisen So the last one is tricky and was going to see discussion on it, so if I change it to what flake8 wants, the code itself fails as it, I believe, is seeing in the array when an element ==true when is true checks the whole array I believe which causes failure, so not sure how to go around this or if this is fine to ignore.

AdamTheisen:
Okay, will look later tonight (if I can stay awake putting the kids down) or tomorrow sometime. I would be okay ignoring buy want to make sure I fully understand

zssherman:
Yeah no worries!

AdamTheisen:
@zssherman could we just do

`idx = np.where(index_list)`

It looks like it gives the same results in testing

AdamTheisen:
This looks good and passes tests, will merge.

PR #359, opened 2021-09-17
FIX: Xarray dataarray fix

zssherman:
Xarray has a new catch error system if defining variables with the dataarray and not .values or .data itself, fix for that error occurring in the unitests now.


zssherman:
forgot to pull upstream like a goofball

PR #360, opened 2021-09-17
FIX: Fix for metpy sounding error.

zssherman:
Error is raised when:
E           metpy.calc.exceptions.InvalidSoundingError:
E                   Pressure does not decrease monotonically in your sounding.
E                   Using scipy.signal.medfilt may fix this.

See if this solves the issue.


zssherman:
I'm wondering if we need a new sounding test file, I'm not sure how to fix this.

AdamTheisen:
@zssherman  I can take a look.  When I've played around with the sonde data for the PBLHT stuff, there's often a time with the same value as previous so we might just need to resample the data

zssherman:
@AdamTheisen Ah gotcha makes sense, and sounds good.

AdamTheisen:
@zssherman I think i have a fix for this, at least it works on my side of things.

zssherman:
Groovy, yeah looks good to me!

AdamTheisen:
@zssherman we should be good to go here. Will look to @zssherman  to check and merge when approved

zssherman:
Looks good merging thanks @AdamTheisen

Issue #361, opened 2021-10-06
CAPE and CIN calculations


avarble:
I saw that you are using MetPy for the CAPE and CIN calculations. Those calculations are incorrect because MetPy uses temperature instead of virtual temperature in the calculations. We reported this here: https://github.com/Unidata/MetPy/issues/2093.

avarble:
@zssherman It should be a quick fix. All you need to do is replace temperature with virtual temperature in the buoyancy calculation for CAPE and CIN.

zssherman:
@avarble sounds good!

zssherman:
@AdamTheisen @avarble From what i'm seeing is, or understanding is virtual temperature is calculated within metpy. So I'm trying to think then would the fix be then to include the virtual temperature calculation within the CAPE and CIN calculations? Not sure if Virtual temperature is typically included in datasets or is needed to be calculated. Based on that, can do changes to reflect that, just trying to figure it out.

zssherman:
@dopplershift Gotcha, yeah if thats the case, I can just used metpys virtual temperature calculation and pass those instead. Thanks for the information!

avarble:
@dopplershift There are certainly different ways of calculating CAPE in terms of whether to keep or unload condensate, whether to include an estimate for fusional latent heating, and of course what you take as your starting parcel properties. But, as far as I know, they all use virtual temperature, not temperature, so using temperature to compute CAPE is incorrect in an absolute sense. It is called a correction to temperature because it is a change in the temperature that corresponds to the water vapor impact on air density (and thus buoyancy).

PR #362, opened 2021-10-07
FIX: Addresses #361, use virtual temperature instead of temperature.

zssherman:


zssherman:
@AdamTheisen So I think I have a working code, unit tests are probably going to change because values will change, however im going to do a pr, and have a review, because the met side of things I want to check with you all, and if it all looks good ill change the unit tests

AdamTheisen:
@zssherman  as expected, it looks like there are some slight differences in the unit test results'

E           Mismatched elements: 5 / 5 (100%)
E           Max absolute difference: 0.36768183
E           Max relative difference: 0.00136254
E            x: array([270.217682, 270.112815, 270.045454, 269.989816, 269.939635])
E            y: array([269.85    , 269.745307, 269.678057, 269.622511, 269.572413])

Meteodan:
All, I found this conversation and the related #361 when trying to get to the bottom of what would be needed to add the virtual temperature correction to MetPy's CAPE calculations. Unfortunately, I don't think the approach you are using here is correct. The problem comes when you are calling MetPy's `surface_based_cape()` function here: https://github.com/ARM-DOE/ACT/pull/362/files#diff-605c6300a0861a7eaad398f848dd5baa1871d6490dcfcd13224c6f6e9c56d0daL196

While definitely the environmental virtual temperature needs to be used in the calculation, what is happening as near as I can tell is that the `surface_based_cape()` function is internally computing the parcel's *uncorrected* temperature profile and using *that* in the CAPE calculation.  To get the correct CAPE with the virtual temperature correction, the parcel's *virtual* temperature profile needs to be computed as well. Another problem is that, since the environmental virtual temp is being passed in in place of the regular temp, then the starting temperature that is being used internally to compute the parcel profile is using the starting virtual temperature instead, while still using the uncorrected dewpoint. This means that the LCL used internally for the CAPE calculation will be wrong (too high) on top of the parcel profile being incorrectly computed.

AdamTheisen:
@Meteodan,  thanks for reaching out and reporting this.  I turned this into an issue (#631) and we will look into it.  Of course we always welcome contributions from others if you have a potential fix in mind!

PR #364, opened 2021-10-15
Sondepbl

AdamTheisen:
over 2 months of test data, the regime classification is about 96% the same as the ARM VAP.  The average height difference the results of ACT and the ARM VAP is about 34m and median of 5m.


AdamTheisen:
@zssherman just pushed one small update but this should be ready for your review when you have time.  Thanks!

Issue #363, opened 2021-10-14
Correctly handle greater than 32 QC tests 


kenkehoe:
There is a file with 32 QC tests defined in the ARM method. When the cleanup is run to convert to CF QC standard the bitpacked number for the 32nd test is causing an overflow issue and is circling around to negative numbers. Need to detect the overflow error and upcast the data type to int64.

I'm working on this now, you can assign to me.

PR #365, opened 2021-10-15
QC Tests data type

kenkehoe:
Fixing issue with setting flag_masks values when converting from ARM to CF QC standard. Turns out ARM uses type int32 in the data file for a QC variable up to 32 tests. But to create the flag masks value need to use uint32. This was causing an error of wrapping around to negative numbers for mask numbers. Solved by using uint32 for flag_masks. Also, updated to correctly upconvert data types when adding more than 32 tests.

Issue #363

### DIFF
--- a/act/tests/test_qc.py
+++ b/act/tests/test_qc.py
@@ -230,11 +230,11 @@ def test_qcfilter2():
     ds_object[var_name].values = data
 
     coef = 1.4
-    ds_object.qcfilter.add_iqr_test(var_name, coef=1.4, test_assessment='Bad', prepend_text='ACT')
+    ds_object.qcfilter.add_iqr_test(var_name, coef=1.4, test_assessment='Bad', prepend_text='arm')
     assert np.sum(ds_object[expected_qc_var_name].values) == 28
     assert ds_object[expected_qc_var_name].attrs['flag_masks'] == [1]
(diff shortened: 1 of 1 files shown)


AdamTheisen:
Unit test coverage decreased a bit more than expected.  We might need to review the unit tests and see if we need to expand them at all.

AdamTheisen:
@kenkehoe I was trying to add in a test to see if I could get better coverage, but I am running into errors.  If I remove flag_values and run

    ds_object.qcfilter.add_test(var_name, index=list(range(0, 20)), test_number=32,
                                test_meaning='Testing flag',
                                test_assessment='Suspect')

The code encounters an error at line 300 in that there's not a flag_masks attribute.  Note, I just added the about to test_qcfilter().

kenkehoe:
OK I'll increase test coverage and look into that error.

AdamTheisen:
@kenkehoe there are going to be additional errors with the tests as it looks like this is changing some fundamentals of the QC.  Is this expected to cause major differences?

kenkehoe:
I'm looking into this now. There was an incorrect option set that is causing issues with some tests.

kenkehoe:
@AdamTheisen  looks like the build process timed out. Can I restart the build process again?

PR #369, opened 2021-11-04
ENH: There's a but in that open_mfdataset defaults to open_dataset if…

AdamTheisen:
… it's one file.  Putting in a check so open_mfdataset keywords aren't passed to open_dataset


jhemedin (approved):
All looks good to me. Accepting and merging

PR #370, opened 2021-11-19
ENH: Updating requires to not include matplotlib 3.5.0

AdamTheisen:
Matplotlib 3.5.0 breaks the sonde plotting so ensuring 3.4.3 or less is used for matplotlib


AdamTheisen:
Going to merge and release a micro

PR #371, opened 2021-12-08
Add in nan

kenkehoe:
I noticed add_in_nan() taking a long time to work. After looking into it I think we were adding too many NaN values for what was needed. The updates will only add in one or one row of NaNs using the build in numpy functions which is much faster. Also, changed to return the data in same type Number or Xarray.DataArray as supplied to the function.


kenkehoe:
Hmmm, this is interesting. I had to change a test comparison value because it was different on my computer. I'll change it back. Interesting that different computers give different answers.

zssherman (approved):

zssherman:
Everything looks good to me, merging.

PR #375, opened 2022-01-05
FIX: An attempted fix to skyfield files.

zssherman:
SKyfield .close method does not fully close the file. Adding a python del to fully delete the file. Can't add load_file in a with statement as skyfield object does not have a __enter__ or __exit__ for this.

### DIFF
--- a/act/utils/geo_utils.py
+++ b/act/utils/geo_utils.py
@@ -67,31 +67,36 @@ def destination_azimuth_distance(lat, lon, az, dist, dist_units='m'):
     return np.degrees(lat2), np.degrees(lon2)
 
 
-def add_solar_variable(obj, latitude=None, longitude=None, solar_angle=0., dawn_dusk=False):
+def add_solar_variable(obj, latitude=None, longitude=None, solar_angle=0.,
+                       dawn_dusk=False):
     """
-    Add variable to the object to denote night (0) or sun (1).  If dawk_dusk is True
-    will also return dawn (2) and dusk (3).  If at a high latitude and there's sun, will
-    label twilight as dawn; if dark{2}, will label twilight as dusk(3).
+    Add variable to the object to denote night (0) or sun (1). If dawk_dusk is
+    True will also return dawn (2) and dusk (3). If at a high latitude and
+    there's sun, will label twilight as dawn; if dark{2}, will label twilight
+    as dusk(3).
 
     Parameters
     ----------
     obj : xarray dataset
         ACT object
     latitude : str
-        Latitude variable name, default will look for matching variables in object
+        Latitude variable name, default will look for matching variables in
+        object.
     longitude : str
-        Longitude variable name, default will look for matching variables in object
+        Longitude variable name, default will look for matching variables in
+        object.
     solar_angle : float
-        Number of degress to use for dawn/dusk calculations
+        Number of degress to use for dawn/dusk calculations.
     dawn_dusk : boolean
-         If set to True, will add values 2 (dawn) and 3 (dusk) to the solar variable
+         If set to True, will add values 2 (dawn) and 3 (dusk) to the solar
+         variable.
 
     Returns
     -------
     obj : xarray dataset
         Xarray object
-    """
 
+    """
     variables = list(obj.keys())
 
     # Get coordinate variables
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
Do you know what additional step Skyfield could take to get the file fully closed? I've never seen a ResourceWarning during testing, and even after just tightening up the `jplephem` tests to try to detect any dangling files, I've not been able to force a ResourceWarning to appear:

https://github.com/brandon-rhodes/python-jplephem

zssherman:
Hi @brandon-rhodes, thanks for the response! When using .close() i was still able to access our file being defined 'planets', making me believe it wasn't fully closed, but I was wrong based on what you stated. Was there a skyfield update recently that worked on closing in previous releases? I'm wondering if on our systems then if we have an outdated skyfield perhaps. @cgodine do you have perhaps a full error output?

brandon-rhodes:
> Was there a skyfield update in the past that worked on closing in previous releases? I'm wondering if on our systems then if we have an outdated skyfield perhaps.

It looks like the `close()` method was added in https://github.com/skyfielders/python-skyfield/commit/61e8e0e9476f87c975fa7cd08df752383b79e634 and would have been released in Skyfield 1.22 on 2020 June 9. The logic it calls in jplephem hasn't been touched for several years.

Separately from the dangling-resource issue your users have run into: would it be helpful if Skyfield’s `SpiceKernel` became a context manager that closed itself? I've never needed that feature myself, but if it would make it easier for your project to work with Skyfield then I could see about adding it for the next version.

zssherman:
Ah gotcha @brandon-rhodes , hmm, so looks like we have the most recent release. Again I appreciate the help with this! I believe that the context manager would be useful! I'm just confused still on why on our end we have the resource warning and don't want to cause more work for yourself if its something on our end. Once cgodine post the full error, it might help give us a clue, but based on what you shared, us using .close() on our end should work.

AdamTheisen:
@zssherman @cgodine can confirm but they are running some multiprocessing so there could be a lot of processes trying to access the same resources at the same time.  Wonder if that is part of it???

cgodine:
Hey @zssherman, sorry for the delay. Here is the full traceback of one of the errors: 

`/home/dqmgr/.conda/envs/dqo-base/lib/python3.8/site-packages/skyfield/functions.py:52: ResourceWarning: unclosed file <_io.BufferedReader name=/home/dqmgr/.conda/envs/dqo-base/lib/python3.8/site-packages/act/utils/conf/de421.bsp>
  return sqrt((xyz * xyz).sum(axis=0))
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/home/dqmgr/.conda/envs/dqo-base/lib/python3.8/site-packages/matplotlib/transforms.py:93: ResourceWarning: unclosed file <_io.BufferedReader name=/home/dqmgr/.conda/envs/dqo-base/lib/python3.8/site-packages/act/utils/conf/de421.bsp>
  self._parents = {}
ResourceWarning: Enable tracemalloc to get the object allocation traceback`

We do run multiprocessing on production, @kenkehoe could probably give you more information on that if needed.

kenkehoe:
@brandon-rhodes we run RedHat Linux 7.9

brandon-rhodes:
Thank you, @kenkehoe!

It occurs to me that I should double-check: @zssherman, did the `del planets` definitely solve the problem and get @cgodine's problem solved? Or was it added to the code but not yet confirmed to be a solution? I just reviewed the conversations but can't find that @cgodine ever reported exactly how the new code behaved on their platform.

zssherman:
@brandon-rhodes The del seemed to delete it locally for me. However with the trace back and mention if multiprocessing I'm wondering as you stated if during the garbage collection it's been detected and the warning is appearing, but appearing multiple times due to the multiple threads and is clogging everything up. We had a similar issue with another package cmac and a cartopy warning when multiprocessing. When I closed I saw the planets variable still, which made me believe it wasn't closing properly and using del deleted it completely, or so I thought. Once you shared the code for closing in skyfield and this discussion I'm thinking it could be as you stated above more then anything.

zssherman:
@brandon-rhodes This is an example of the close and still being able to access the saved variable:
Screenshot from 2022-01-07 12-32-34
skyfield version 1.39

I wanted to double check because I was running some stuff in Jupyter notebook and wanted to make sure if something goofy was going on using that.

kenkehoe:
@zssherman  I don't think the file is actually open after a call to .close() . I ran the same code you did above and in another window used the linux command to see how many processes were connected to the file (lsof -t de421.bsp | wc -w). While the planets object is open I see one connection to the file. After running planets.close() I do not see any connections to the file. I think the access to planets.codes is from the values stored in memory. I don't think you need to make a release right now. I'll keep looking into this.

brandon-rhodes:
@kenkehoe — Thanks for jumping in so quickly with that observation!

@zssherman — The `.codes` are a small set of Python integers, read in from an ephemeris segment list when the ephemeris is first opened. It hadn't occurred to me to empty it out once the ephemeris was closed, but I think it might be much clearer to users if it did! I could also empty the `.segments` list. Would that improvement to Skyfield have given you more confidence that the file was closed?

zssherman:
@brandon-rhodes @kenkehoe Ah that makes sense! Thanks for clearing that up! Nah I don't think anything needs to be changed, I just misunderstood and should have check the connections. Mistake on my part, I'll close this for now. Thank you both for the information and clarification!

PR #381, opened 2022-01-11
Global attribute qc

kenkehoe:
There is an issue when ARM data files define the QC tests in the global attributes but not all the tests are used on all the variables. When converting from globally defined tests to variable defined tests need to check if the valid_min, valid_max or valid_delta attributes are defined for each variable to determine if the test was performed. This PR fixes this issue. Also, fixing some pytest stuff.


zssherman (approved):
The changes look good, the unit tests seemed to be failing because of a tolerance change with python. The more significant digits, seems to be minor difference between the comparison of the tests. Might need to adjust the atol, I can try to tackle this soon.

kenkehoe:
Yeah I noticed this in the past, but the values ended up being correct for tests here. I'll look into this and see if I can resolve.

zssherman:
@kenkehoe Nice! Thanks for doing that! Everything looks good to me, merging.

kenkehoe:
@zssherman can you create conda releases? I have a few merged code updates I'd like to install on our production system.

zssherman:
@kenkehoe I can do that, i'll get working on that shortly here.

zssherman:
@kenkehoe The conda forge PR has been merged, in a few minutes or so, 1.1.2 should be available on conda-forge.

PR #386, opened 2022-01-20
Matplotlib Marker keyword

kenkehoe:
The marker keyword is hardcoded into a plotting call, so specifying the marker in an ACT plotting call is causing a conflict. This PR puts the marker in the kwargs dict so we can have a default and can override default. Found a few other instances of keywords being set in the call so fixed those as well.


kenkehoe:
I don't think the tests ran for this. Can we try them again?

zssherman:
I believe because we are adding the ax.plot for lines, the images do not match for the unit tests and that why its failing.

kenkehoe:
@zssherman  OK I see the issue. I can drop the lines but I'm confused why the colors have changed. Any ideas? I can save an updated image if that is needed.

AdamTheisen:
@kenkehoe I think it might be that your kwargs['edgecolors'] has an s and the keywords were just edgecolor

AdamTheisen:
I'll take a look at this.  Curious why we would need to set linestyle to 'None' on this

display.plot('backscatter', force_line_plot=True, linestyle='None')

kenkehoe:
I think the default is to draw the line for Matplotlib. The baseline image does not have lines. The strange thing I'm seeing is that the colors are different. I wonder if the limit for image comparison is not tight enough since my colors are different than the baseline, but tests on my laptop would pass. Most of the image is just dots so connecting the lines is not obvious with so many points.

kenkehoe:
Now I see there are two different keywords: edgecolor and edgcolors. I may need to double check I managed both correctly.

AdamTheisen:
@kenkehoe I changed to edgecolor and it seems to work without the change to the test_plotting script

kenkehoe:
I think edgecolors was the correct keyword the pcolormesh call. But it also looks like we are setting the keyword to the standard default so maybe we just remove the keyword and let the kwargs handle it?

zssherman:
Yeah if its already the default, is edgecolor used often? Well more is it changed often? If not it would probably be better to leave it to the kwargs

kenkehoe:
@AdamTheisen  you OK with removing the keyword from the function call and expect people to just add it via the kwargs method?

AdamTheisen:
@kenkehoe as long as it doesn't change the default behavior I would be ok with that.

kenkehoe:
OK I think my update of edgecolor in the broken_bahr was not quite right. I've updated to be more like the original but add keyword documentation that was missing.

AdamTheisen:
@kenkehoe @zssherman I might have misunderstood.  Was the intent to sent edgecolors to the plot function through kwargs or still have it as a keyword to the plot function and then pass to plot/pcolormesh in kwargs?

kenkehoe:
I think we are not using the default for edgecolors so to set it in a way where it can be overwritten by a keyword I have the keyword set in the normal way, but passed to pcolormesch through kwargs dictionary.

AdamTheisen:
@kenkehoe  thanks for jogging my memory there.  Will approve and merge unless @zssherman has anything else to address.

AdamTheisen (approved):

PR #388, opened 2022-02-02
Fix flag_values set as a single string

kenkehoe:
Catching cases when flag_values or flag_masks are incorrectly set as strings. Sometimes we have incorrectly formatted files. This update will catch and fix those instances. Such is the life in ARM.

### DIFF
--- a/act/qc/clean.py
+++ b/act/qc/clean.py
@@ -176,8 +176,6 @@ class CleanDataset(object):
                 try:
                     if self._obj[var].attrs['units'] not in ['1', 'unitless', '', ' ']:
                         continue
-#                    self._obj[var].attrs['valid_range']
-#                    continue
                 except KeyError:
                     pass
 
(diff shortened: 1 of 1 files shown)


kenkehoe:
So I have a failure of the ACT/act/tests/test_retrievals.py::test_get_stability_indices() function with the metpy units.

ValueError: `mixing_ratio_from_relative_humidity` given arguments with incorrect units: `relative_humidity` requires "[dimensionless]" but given "hectopascal", `pressure` requires "[pressure]" but given "percent"

I tried to fix it to test correctly on my laptop but the error makes no sense. Does anyone else have this issue. Since it passed I assume it's OK? But, this should not stop the PR.

AdamTheisen:
@kenkehoe  I've seen that before on my end but I think it was related to the version of metpy i had installed.  However, it doesn't seem to be an issue here since all jobs pass.  Code looks good and will merge.

Issue #387, opened 2022-02-01
numpy dtype error when plotting time


gg2:
While testing ACT to fetch arbitrary plot data, I'm running it against some local VAP data. For most variables it plots successfully; but if I try to plot the variable 'time', I get the following error:
"
The DType <class 'numpy.dtype[datetime64]'> could not be promoted by <class 'numpy.dtype[float64]'>. This means that no common DType exists for the given inputs. For example they cannot be stored in a single array unless the dtype is `object`. The full list of DTypes is: (<class 'numpy.dtype[datetime64]'>, <class 'numpy.dtype[float64]'>)
"

'time' is specified as double in this data.
At the moment, I'm running this on Python 3.7.9 on Windows 10.
```
    nc_data = act.io.armfiles.read_netcdf(files_list)
    ts_display = act.plotting.TimeSeriesDisplay(nc_data, figsize=(19,11), subplot_shape=(2,))
    ts_display.plot(var_name, subplot_index=(0,))
```

`files_list` is a list of filenames gathered per specified date ranges.

If desired and unable to replicate the error, I can point you at the data on ARM development servers.

AdamTheisen:
Thanks for reporting @gg2 ! We will take a look soon and start some testing

AdamTheisen:
Okay, this looks like our handling of the y-axis range needs to be improved to handle these types of cases.  I will see what I can do

AdamTheisen:
@gg2 there was a little more to this since by default xarray and ACT convert time to a datetime.  I have made a couple updates to the code that I'll get a PR going on.  Once that's in place, if you want to plot the actual 'time' double values, there are some keywords you will need to set when reading the data in.

`nc_data = act.io.armfiles.read_netcdf(files,decode_times=False, cftime_to_datetime64=False)
ts_display = act.plotting.TimeSeriesDisplay(nc_data)
ts_display.plot('time')
`
or you could just leave the read portion as 

`nc_data = act.io.armfiles.read_netcdf(files)`

and it will plot the data as datetime64s.  Plots of the double and datetime64 values are attached.  PR coming shortly.  Let me know if this resolves the problem you were having.

![time_plot_double](https://user-images.githubusercontent.com/13421074/152275951-c84cbad1-9e51-49db-8c6f-ff9cb8cd2baf.png)
![time_plot_datetime64](https://user-images.githubusercontent.com/13421074/152275952-0ee86ecf-5c8d-4b33-a02d-b91e9a86ef17.png)

[...]

PR #389, opened 2022-02-03
Time plot

AdamTheisen:
Updating the timeseries code to allow for the plotting of time as a double or a datetime64


AdamTheisen:
In response to #387

zssherman:
Looks good to me!

AdamTheisen:
@zssherman the coverage decreased more than expected so I will try and add a quick unit test for plotting out the time

PR #390, opened 2022-02-15
Fixing logic from a return to a continue

kenkehoe:
The cleanup QC method has a logic error where it returns inside a loop when it should continue. Additional variables are not being processed.


zssherman (approved):
Looks good to me, tests still running.

PR #393, opened 2022-02-18
ENH: Updating for the latest crop information/returns

AdamTheisen:


zssherman:
@AdamTheisen Will these be likely to change again, or was it more of a naming convention change?

zssherman:
@AdamTheisen Ah gotcha makes sense! Everything looks good to me.

Issue #391, opened 2022-02-18
Adding qc tests becomes very slow for datasets with a large number of variables


maxwellevin:
I'm supporting a `tsdat` data user who has with a dataset with 300+ variables in that they want to apply some qc to. Tsdat uses the `act` `qcfilter.add_test(...)` dataset accessor under the hood to record the results of our quality checks and we're finding that for large datasets this method can quickly get prohibitively expensive. At first I thought it was because we had to generate and attach new `DataArray` objects to the dataset for each quality check, but I noticed no speedup at all for `ds.qcfilter.add_test(...)` when I created the qc variables in advance.

Here's a basic example I created to illustrate what I'm seeing:

```python
def run_example(n_variables, n_samples=100):
    time = pd.date_range(
        start="2022-02-17 00:00:00",
        end="2022-02-18 00:00:00",
        periods=n_samples + 1,
    )

    # Create data variables and initialize QC variables to 0
    noisy_data_mapping = {f"data_var_{i}": np.random.random(time.shape) for i in range(n_variables)}
    qc_vars = {f"qc_data_var_{i}": np.zeros(time.shape) for i in range(n_variables)}
    qc_vars.update(noisy_data_mapping)

    ds = xr.Dataset(
        data_vars={name: ("time", data) for name, data in noisy_data_mapping.items()},
        coords={"time": time},
    )


    start = datetime.datetime.now()
    for name, var in noisy_data_mapping.items():
        failed_qc = var > 0.75  # Consider data above 0.75 as bad. Negligible time here.
        ds.qcfilter.add_test(
            name,
            index=failed_qc,
            test_number=1,
            test_meaning="Value above threshold.",
            test_assessment="Bad",
        )
    end = datetime.datetime.now()
    print(f"Time elapsed = {(end-start)}")
```

When I run this code on my laptop for various values of `n_variables` I got the following results:
```
n_variables = 25 --> 0.5 seconds
n_variables = 50 --> 3.8 seconds
n_variables = 100 --> 35 seconds
n_variables = 200 --> 643 seconds
```

Roughly speaking, doubling the number of variables in the dataset seems to result in a tenfold increase in the time taken to add qc tests to that dataset, though the jump from `n_varables = 100` to `n_variables = 200` is more like a 20x increase. Oof.

maxwellevin:
Using a profiler sheds even more light on what's going on. Here's the results of the profiler for `n_variables = 100`:

[...]

PR #394, opened 2022-02-19
QC processing time

kenkehoe:
Issue #391 pointed out a performance issue with setting a new QC test. Turns out there was a call to .clean.cleanup() in the add_test() method via check_for_ancillary_qc() that was unnecessary and causing a significant number of loops to be executed. The default has been changed so clean.cleanup() is not run. Also, updated the way the QC variable names are extracted from the object to be more efficient and more accurate.

To ensure this type of thing does not happen again with a change to the QC code, I've added a test of the performance of the add_test() method. It uses wall clock time so it's not the best methodology, but should highlight significant performance decreases.

Overall this is a significant performance boost with a 60 fold decrease in computation time.

### DIFF
--- a/act/qc/clean.py
+++ b/act/qc/clean.py
@@ -40,38 +40,43 @@ class CleanDataset(object):
             A list of strings containing the name of each variable.
 
         """
-        variables = []
 
         # Will need to find all historical cases and add to list
-        qc_dict = {'description':
-                   ["See global attributes for individual.+bit descriptions.",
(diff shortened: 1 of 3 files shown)


zssherman:
Everything looks good to me.

zssherman (approved):

AdamTheisen (approved):

AdamTheisen:
Looks good, will merge in.  @zssherman let's plan to do a version release when we're back online this week.

PR #404, opened 2022-03-02
ADD: New PySP2 wrapper for ACT

rcjackson:
This is a wrapper around the main PySP2 functions for ACT so that ACT can read and process SP2 data. 

PySP2 is now moved to the ARM-DOE repo with docs up:
https://arm-doe.github.io/PySP2/

In addition, PySP2 is now on PyPI and will soon be on conda-forge.


zssherman:
Thanks for the PR! I'll take a look here shortly.

zssherman:
Tests pass and code looks fine, I would just remove the new lines that are appearing in the __init__ file change.

zssherman:
Yeah looks good, looks like travis isn't running, will have to move to github actions soon.

act/qc/sp2.py
```diff
@@ -0,0 +1,100 @@
...
+        c0Scat2, c1Scat2, c2Scat2: floats
+            Calibration scattering coefficients for ch4
+        densitySO4, BC: float
+            Density of SO4, BC
+        tempSTP, presSTP: float
+            Temperature [Kelvin] and pressure [hPa] at STP.
+
+        """
+        def __init__(self, cal_file_name=None):
+            super().__init__(cal_file)
```
zssherman, line 63: Is this suppose to be cal_file_name as well? wondering if that is causing the error.

zssherman: @rcjackson

rcjackson: I've fixed the name mismatch here.

PR #406, opened 2022-03-03
CI: Adding Github actions tests for linux

zssherman:
Will see if this works...


zssherman:
If this does work, i can add testing for mac and maybe windows as well.

AdamTheisen:
@zssherman we will have to do the documentation build as well right?  And yes, let's look to wider testing as well, especially with Windows

mgrover1:
@zssherman can you add `flake8` to the environment?

zssherman:
@mgrover1 Yeah sounds good, I'll give that a shot

zssherman:
Still no luck sadly @mgrover1

zssherman:
Side note once this is figured out, I plan on squashing these all into 1 commit to clean it up.

zssherman:
rerunning with bobbys fix to his PR.

zssherman:
Groovy so the recommendation by Joe O'Brien and max as well as a fix for Bobby's code should get this running again.

mgrover1:
@zssherman I would change the line where you run pytest to ensure you are using your conda env

```bash
python -m pytest
```

mgrover1:
It looks like it's having a tough time getting `MutableMapping` - I think it needs to be imported from `collections.abc`

```python
from collections.abc import MutableMapping
```

zssherman:
@mgrover1 yeah it seems all three environments used python 3.10 I think because I removed the python line for the environment used line. Pint i don't think has fixed their deprecations for python 3.10 which is why i think it failed it for all three. If 3.8 and 3.9 pass now, think 10 we can't tests till pint updates.

zssherman:
Yep environments are now working correctly just metpy causing issues, which I believe we were pip installing to solve. So moving to back to that.

jrobrien91:
> Just realized is everyone receiving a million emails as I try to fix this?

Yes, but I'm learning a lot so thanks!

mgrover1:
@zssherman the only solution I could find is here https://stackoverflow.com/questions/48614440/cannot-cast-array-data-from-dtypem8ns-to-dtypefloat64-according-to-th

zssherman:
Why this is so strange is because an exact replica on my local machine these pass. Adam mentioned this already being fixed. Pytest seems to be called from within the act env in the actions run. So I have no idea how this is failing here.

zssherman:
Tests now pass! adding reviewers.

AdamTheisen (approved):
Looks good!  Will be great to move away from Travis-CI

mgrover1 (changes requested):
Generally looks good! What is the reason for having two environment files, with one in the directory with the github action?

zssherman:
@mgrover1 the one in github actions I have in place for the ci. The one in main directory was for users to create environments with no optional dependencies

mgrover1:
@zssherman what are your thoughts about adding a `ci` directory? That way if we have future things to add there, we can, and its more transparent than hiding it in the hidden `.github` directory?

zssherman:
I think that's a good idea. So add the environment file to the ci folder?

mgrover1:
yeah!!

zssherman:
we have a continous integration folder so what I can do is more it there and remove the travis stuff

zssherman:
@mgrover1 I moved the environment file to the continous_integration folder as well as removed some of the travis files. I didn't remove travis completely yet as the docs are built that way. Once we switch to docbuild here as well in a different PR can remove travis competely.

mgrover1 (approved):
Thanks for making those changes @zssherman ! Looks great!

PR #411, opened 2022-03-08
Update branch to master not main

mgrover1:
I noticed the docs did not start to build since I used main instead of master


AdamTheisen:
Otherwise checks look good!

PR #420, opened 2022-03-18
Add new google analytics ID

mgrover1:
Update to use the new google analytics ID that won't be replaced in a year


AdamTheisen:
Looks good, will merge in

PR #421, opened 2022-03-21
Fix Google Analytics ID bug

mgrover1:
After checking the dashboard, it looks like the analytics aren't working.. this fixes that bug


zssherman (approved):
Looks good! Feel free to merge once the tests pass.

PR #423, opened 2022-03-22
ENH: Updating all the examples to better display images and fix bugs

AdamTheisen:


AdamTheisen:
@mgrover1 I might want to get your insight here.  The python code is not actually seeing the environment variables set in Github. https://github.com/ARM-DOE/ACT/runs/5648965745?check_suite_focus=true

act/tests/test_discovery.py
```diff
@@ -55,17 +54,18 @@ def test_get_armfile():
     if not os.path.isdir(os.getcwd() + '/data/'):
         os.makedirs(os.getcwd() + '/data/')
 
-    uname = os.getenv('ARM_USERNAME')
-    token = os.getenv('ARM_PASSWORD')
+    # Place your username and token here
+    username = os.environ['ARM_USERNAME']
```
mgrover1, line 58: Can you change this back to `os.getenv('variable')`? For example,

```python
var = os.getenv('SAMPLE_SECRET')
```

I tried this out in this repo https://github.com/mgrover1/test-actions, it seemed to work (although it hid the secret from printing  - which is to be expected)

Sample output
```bash
Run python print_env_var.py
[7](https://github.com/mgrover1/test-actions/runs/5649861907?check_suite_focus=true#step:4:7)
***
```

mgrover1 (approved):
Looks great - thanks @AdamTheisen - I like the removal of "Examples..." from the title of each of those updated examples, and the addition of the discovery test.

PR #428, opened 2022-03-23
Coveralls and pytest updates

AdamTheisen:
Adding coveralls back in and updating pytest to run pytest --mpl --cov=act/ --cov-config=.coveragerc


AdamTheisen:
Close out to see if I can get rid of all those commits

PR #429, opened 2022-03-23
ADD: Adding coveralls back in, updating pytest and flake8 as well

AdamTheisen:
Pytest and flake8 requirements were updated with the switch to github actions which is causing us to miss a few things I think on PRs.  Trying to bring back to original standards.


zssherman (approved):

AdamTheisen:
Everything passed so will merge in

PR #435, opened 2022-03-25
FIX circular import for PySP2

rcjackson:
There was a circular import occurring in the SP2 wrapper that was causing PySP2 to break. I've removed the circular import and moved the full contents of the offending data structure into ACT as its own unique class with the same attributes. PySP2 tests ran on my laptop after this happened.


zssherman:
Not sure why the coverage decrease, I'm wondering if how pytest behaves with skips. Would changing it to:
@pytest.mark.skipif(not PYSP2_AVAILABLE,
                                 reason="PySP2 is not installed.")
above each test function? Wondering if having it within the test function is causing issues.

rcjackson:
Did that fix the issue @zssherman ? I added those lines.

AdamTheisen:
@rcjackson looks like it's failing on some flake8 now 

./act/tests/test_retrievals.py:243:1: E128 continuation line under-indented for visual indent
./act/tests/test_retrievals.py:254:1: E128 continuation line under-indented for visual indent

AdamTheisen:
Something still strange going on here in that it doesn't appear to actually import pysp2 correctly.

https://coveralls.io/builds/47707885/source?filename=act%2Fqc%2Fsp2.py

AdamTheisen:
@zssherman @mgrover1 can you check this error out? It's unrelated to the PR and looks to be a problem with PyART???

rcjackson:
I can't tell if it's working or not. Looks like PyART is breaking the whole thing.

mgrover1:
@AdamTheisen I think it has to do with the cython/numpy versions... ive been looking into this along with the CyLP stuff today...

AdamTheisen:
This actually just failed with the latest PR I just merged that was passing so something changed in the past day.  I'll open an Issue and more this conversation there @mgrover1

zssherman:
Rerunning all tests with pinning PR now in.

zssherman:
Are we able to test all the self. definitions? create a object then assert the values equal what those are set at?

mgrover1:
@zssherman I think it's running into an error saying that PySP2 isn't installed... can we add that to the environment?

AdamTheisen:
@mgrover1 @zssherman looking at the raw logs, it looks like an older version of pysp2 is being installed

```
Successfully installed .... pysp2-1.0.3 ....
```

I only see version 1.0 on github but on pip the latest is 1.0.7.  Could we try and pin that version?

AdamTheisen:
@mgrover1 I was looking at the raw logs and if you search for pysp2 you can see it's actually collecting a number of different versions of pysp2 but it says it only successfully installed 1.0.3.

The missing files are due to the older version of pysp2 I believe in that we import those test files from pysp2

```
2022-04-05T14:03:47.5947565Z Collecting pysp2
2022-04-05T14:03:47.5948031Z   Downloading pysp2-1.0.7-py3-none-any.whl (14.0 MB)
2022-04-05T14:03:47.5967070Z Collecting pysp2
2022-04-05T14:03:47.5967498Z   Downloading pysp2-1.0.6-py3-none-any.whl (14.0 MB)
2022-04-05T14:03:47.5968117Z      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 14.0/14.0 MB 100.8 MB/s eta 0:00:00
2022-04-05T14:03:47.5968691Z   Downloading pysp2-1.0.5-py3-none-any.whl (14.0 MB)
2022-04-05T14:03:47.5969222Z      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 14.0/14.0 MB 78.9 MB/s eta 0:00:00
2022-04-05T14:03:47.5969739Z   Downloading pysp2-1.0.4-py3-none-any.whl (14.0 MB)
2022-04-05T14:03:47.5970340Z      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 14.0/14.0 MB 116.2 MB/s eta 0:00:00
2022-04-05T14:03:47.5970896Z   Downloading pysp2-1.0.3-py3-none-any.whl (14.0 MB)
2022-04-05T14:03:47.5971425Z      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 14.0/14.0 MB 120.5 MB/s eta 0:00:00

2022-04-05T14:03:47.6006501Z Successfully installed appdirs-1.4.4 arm-pyart-1.11.8 charset-normalizer-2.0.12 metpy-1.2.0 mpl2nc-1.3.5 numpy-1.21.4 oldest-supported-numpy-2022.3.27 pint-0.19 pooch-1.6.0 pysp2-1.0.3 requests-2.27.1
```

mgrover1:
Thanks @AdamTheisen - I'll add a requirement that installs a more recent version

mgrover1:
It's failing on all of the Windows builds since the year 1904 is before 1970... here is [a thread on this issue](https://stackoverflow.com/questions/59199985/why-is-datetimes-timestamp-method-returning-oserror-errno-22-invalid-a)

>       base_time = datetime.datetime(1904, 1, 1).timestamp()
[375](https://github.com/ARM-DOE/ACT/runs/5838122438?check_suite_focus=true#step:6:375)
E       OSError: [Errno 22] Invalid argument

@rcjackson have you tested this PySP2 on windows before? Do you know a good way around this?

mgrover1:
I submitted a fix upstream in [PySP2](https://github.com/ARM-DOE/PySP2/pull/28) which should fix this issue - but the problem becomes how we deal with getting the updated PySP2 in here since we are stuck with numpy version conflicts with the recent versions...

zssherman:
Would it be better to just pin to the most recent numpy instead of oldest numpy? Maybe just windows 10 failure and can go from there.

mgrover1:
> Would it be better to just pin to the most recent numpy instead of oldest numpy? Maybe just windows 10 failure and can go from there.

I think it might be better to figure out a solution upstream, and cut a new release of PySP2 with less strict dependencies... thoughts @rcjackson ?

AdamTheisen:
Looks like the tolerance is slightly high.  We've had similar issues with other scripts in the the code works fine but on the CI system produced varied results and had to bump up the tolerance a bit.  If we want to increase from 30 to 35  or more I think that would be okay if that would resolve these pytest errors

zssherman (approved):

AdamTheisen:
I think @mgrover1 should do the honors of merging this with all the work he put into troubleshooting!

Issue #372, opened 2021-12-14
SkewT Plots not working when Pressure does not Decrease Monotonically


ajsockol:
I'm getting this error when trying to plot skewT diagrams: 
![Screen Shot 2021-12-14 at 10 45 18 AM](https://user-images.githubusercontent.com/49569746/146041691-333470fc-9363-49ec-ae36-4fbd9b939e2d.png)

Looking at skewtdisplay.py, it seems there *is* a fix in place for this issue, but it still raises an invalid sounding error when it tries to implement the fix. Maybe we just need to implement some code to remove duplicate pressure levels? Perhaps by removing any points where the difference between two pressure levels is equal to zero?

AdamTheisen:
Thanks @ajsockol !  I've ran into this before with the PBL calculations in the calculate_pbl_liu_liang retrieval code.  I had to do some smoothing to the pressure ahead of time.  The plotting might not be an issue if there are NaNs but we'll have to test that.

kenkehoe:
I was able to fix this in the DQ Office code by looking for differences that are greater than or equal to zero for pressure and resample the Dataset to exclude those values. Should we look into implementing this logic for Sonde Skewt plotting? The other option is to not extract the good values but set the issue causing Pressure values to NaN in a copy of the Dataset.

AdamTheisen:
@kenkehoe with the samples so close together this is what I did in the pbl height calculations

`ds[pressure] = ds[pressure].rolling(time=smooth_height, min_periods=1, center=True).mean()`

if that didn't work, then I increased the smoothing

`ds[pressure] = (ds[pressure].rolling(time=smooth_height + 4, min_periods=2, center=True).mean())`

kenkehoe:
Is this still an issue @AdamTheisen ?

AdamTheisen:
Yes, should have closed out with #436

PR #436, opened 2022-03-25
ENH: Fix for when pressure values are the same or decrease

AdamTheisen:
Tested this on the problem files noted in #372 and looks to work.  Will want @kenkehoe and @ajsockol to take a look as well.


zssherman (approved):

zssherman:
Looks good to me @AdamTheisen!

kenkehoe:
@AdamTheisen  I think this fix will work.

PR #441, opened 2022-03-31
Apply dqr

kenkehoe:
Discovered a problem when applying DQRs to scalar variables. Adding a try/except to hand this case.


mgrover1 (approved):
Looks great! Thanks @kenkehoe

PR #447, opened 2022-04-12
Adding logic to ensure bit wise comparisons and bit flipping is done with integers

### LINKED ISSUE #446: Error when clean routine tries to remove unused bits
When running the script below on the ARM ADC systems, an error occurs when trying to add a DQR to the QC which stems from trying to remove unused bits from those QC variables

```
  File “/home/kehoe/dev/dq/lib/python/ACT/act/qc/qcfilter.py”, line 1065, in unset_bit
    array = array & ~(1 << bit_number - 1)
TypeError: ufunc ‘bitwise_and’ not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ‘’safe’'
```

```
import act
import glob
import numpy as np

ds = 'sgpmetE13.b1'
site = ds[0:3]
files = glob.glob('/data/archive/'+site+'/'+ds+'/'+ds+'.2010*cdf')
years = [f.split('.')[-3][0:4] for f in files]
years = np.unique(years)

for y in years:
    files = glob.glob('/data/archive/'+site+'/'+ds+'/'+ds+'.'+y+'*cdf')
    files.sort()
    obj = act.io.armfiles.read_netcdf(files)
    obj = act.qc.arm.add_dqr_to_qc(obj, variable='temp_mean')
    obj = obj.where(obj['qc_temp_mean'] == 0)
    obj = obj.resample(time='M').mean()
    for i in range(len(obj['time'].values)):
        print(','.join([str(obj['time'].values[i]), str(obj['temp_mean'].values[i])]))
    obj.close()
```

kenkehoe:
When a mean or some other filter is applied to the dataset all variables are converted to data type float. This will break the QC as it requires those variables to be integers. This adds logic to ensure the QC variable is type int when asked to add or remove (set/unset) a test.

This should resolve #446


kenkehoe:
@AdamTheisen  well I'm not sure what to do next. I've updated the baseline plot using the interactive plot window and savefig() method. Both are not making a baseline plot that satisfies the testing. Do you have a suggestion on what to do next? As far as I can see my updated code is correct.

AdamTheisen:
@kenkehoe this is the same issue as the other PR. We might have to look at raising the threshold difference. @mgrover1  thoughts?

mgrover1:
@AdamTheisen currently taking a look at this

mgrover1:
Which it looks like you did update the baseline image - I'll look into this bit more.

mgrover1:
@AdamTheisen and @kenkehoe - it looks like the current CI workflow was installing ACT from the released version since it was a part of the dependencies (PySP2). I pushed a commit that changes

```bash
pip install -e .
```

to

```bash
python -m pip install -e . --no-deps --force-reinstall
```

which upon running the installation, returns

```bash
Installing collected packages: act-atmos
[21](https://github.com/ARM-DOE/ACT/runs/6009341451?check_suite_focus=true#step:4:21)
  Attempting uninstall: act-atmos
[22](https://github.com/ARM-DOE/ACT/runs/6009341451?check_suite_focus=true#step:4:22)
    Found existing installation: act-atmos 1.1.3
[23](https://github.com/ARM-DOE/ACT/runs/6009341451?check_suite_focus=true#step:4:23)
    Uninstalling act-atmos-1.1.3:
[24](https://github.com/ARM-DOE/ACT/runs/6009341451?check_suite_focus=true#step:4:24)
      Successfully uninstalled act-atmos-1.1.3
[25](https://github.com/ARM-DOE/ACT/runs/6009341451?check_suite_focus=true#step:4:25)
  Running setup.py develop for act-atmos
[26](https://github.com/ARM-DOE/ACT/runs/6009341451?check_suite_focus=true#step:4:26)
Successfully installed act-atmos
```

indicating that it uninstalled the released version, and installed the local version! The tests are still running - but this **should** solve the issue, using the newly generated images.

kenkehoe:
@mgrover1  I did not specify the background, just the default. Do we need to specify that?

mgrover1:
@kenkehoe the tests are passing now - I removed the Python 3.10 tests right now, since we are having issues with Matplotlib on Windows with Python 3.10.

mgrover1 (approved):
Tests are passing, code looks great! Thanks @kenkehoe

Issue #448, opened 2022-04-20
ARM Webservice Only works with Dates with -


AdamTheisen:
We should add some more robust processing to the get_armlive download_data module.  When passing in an ARM formatted date (20220101) it does not work and expects 2022-01-01.  We should add some catches for that and additional catches for "not real" dates like 2022-04-31 so that we use the nearest date.

zssherman:
I'll take a shot at this.

mgrover1:
@AdamTheisen what if we took in a date time instead of a string?

kenkehoe:
I think taking in a datetime as one of the options is a good idea, but would prefer we not limit it to just a datetime. I think using dateparser if string, use datetime.datetime or numpy.datetime64 if that is used.

AdamTheisen:
#449  has resolved the initially reported issue

PR #449, opened 2022-04-20
ENH: Adding date parser to get_armfiles.

zssherman:
@AdamTheisen Is this kinda what you had in mind for #448 ? Any of these formats are accepted and then turned into the proper string for the query. If so I can add unittests as well.

### DIFF
--- a/act/discovery/get_armfiles.py
+++ b/act/discovery/get_armfiles.py
@@ -4,7 +4,6 @@ Script for downloading data from ARM's Live Data Webservice
 """
 
 import argparse
-from datetime import datetime
 import json
 import os
 import sys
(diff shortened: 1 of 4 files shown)


kenkehoe:
I agree generalizing this date parsing would be fantastic. Could we add the date_parser() function to the datetime_utils.py file and also add a keyword option for setting the returned value. Maybe pass in the expected value to be passed into strftime so we can use it elsewhere. I believe the DQR function expects YYYYmmdd format.

zssherman:
Did some changes and believe added parameters for above mentioned. Let me know if it looks ok or if I understood correctly and I'll add some unittests.

AdamTheisen:
I think that looks good @zssherman

zssherman:
Groovy, I'll add the unitests to this PR later today.

mgrover1:
@zssherman feel free to merge when you think it's ready!

kenkehoe:
For consistency should we also have a datetime_to_datetime64 function to complement datetime64_to_datetime?

PR #450, opened 2022-04-22
Updated to set data as Dask array back in Dataset if data is originally stored as Dask array

kenkehoe:
Noticed this method was not storing the data back in the Dataset in the same format as was retrieved. This will check if data is set as Dask array and set data to same data type. Also, added some more tests.

### DIFF
--- a/act/qc/qcfilter.py
+++ b/act/qc/qcfilter.py
@@ -942,9 +942,9 @@ class QCFilter(qctests.QCTests, comparison_tests.QCTests):
                 ds.qcfilter.datafilter(rm_assessments="Bad")
                 ds_2 = ds.mean()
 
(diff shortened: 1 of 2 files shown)


kenkehoe:
Also, once this PR is merged can we make a release to provide access to some recent updates. @AdamTheisen

AdamTheisen:
Sounds good @kenkehoe requesting a review from @zssherman and then maybe after he can do a new release and work with @mgrover1 on getting familiar with our process

zssherman (approved):

zssherman:
Sounds good @AdamTheisen !

mgrover1 (approved):
Looks great!

PR #456, opened 2022-04-29
Added keyword to apply clean.cleanup() method instead of requiring the method call after reading data.

kenkehoe:
Also fixed an issue with using pathlib.Path().

### DIFF
--- a/act/io/armfiles.py
+++ b/act/io/armfiles.py
@@ -10,7 +10,7 @@ import json
 import re
 import urllib
 import warnings
-from pathlib import Path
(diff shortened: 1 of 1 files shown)


mgrover1 (commented):
@kenkehoe what is the reason for using `PosixPath` here instead of the more generic `Path`?

mgrover1 (approved):
Thanks @kenkehoe for putting this together!

mgrover1:
@kenkehoe is this good to merge?

kenkehoe:
Yes I believe so @mgrover1

PR #461, opened 2022-05-10
Adding BSRN QC tests

kenkehoe:
The BSRN group have a set of standard tests they use for evaluating data. The tests were written by Chuck Long and Ellsworth Dutton, both long time ARM people. These tests should work well with ARM and any other radiation data sets.


kenkehoe:
@AdamTheisen  I added tests for these new methods but I did not add any examples. Do you think we need to add some example code? We may need to highlight that the BSRN codes are part of ACT now.

AdamTheisen:
@kenkehoe I think examples would be great but could always come as a second PR if you want.  It would be great to advertise this addition

kenkehoe:
OK I'll add some examples.

AdamTheisen (approved):

PR #463, opened 2022-05-12
Fixing issue setting data to Dask array in datafilter

kenkehoe:
Turns out the code was defaulting to set the data back into the Dataset
as a numpy masked array. I think this was done to preserve integer
data type. But when setting the data back as a Dask array this does
not work correctly. Also, not sure the xarray methods would work
correctly anyway so changed to set as Dask array with NaN values
or Numpy array with NaN values. Also added a check to ensure the
needed keywords are set.


zssherman (approved):
Makes sense to me Ken, changes look fine to me.

PR #465, opened 2022-05-13
Documentation update

kenkehoe:
I noticed that the documentation in QC area was not consistent and lacking in examples. I've updated much of the documentation.

Discovered a better way to parse the bits using numpy so implemented that. Left old code for now for reference.

Discovered a keyword for applying DQRs was not actually used. It is not needed. Took it out.

No rush to integrate this. I plan to do other PRs for documentation updates to make the process more manageable.

PR #466, opened 2022-05-19
Improved documentation and new atmospheric pressure test.

kenkehoe:
Updated documentation within the methods and function calls.

Added a new test to use MetPy pressure derived from altitude
to check if values are within expected range.

Fixed issue with writing to file in CF format and reading file
back into Dataset. There was an issue with the flag_* variable
attributes not being in a list or array as expected when only one
value exists.

Added additional tests for the new atmospheric pressure testing
function.

I tried to keep the updates in this separate from PR #465 but I may have some of those as well. Does it matter?


mgrover1 (approved):
Great! Thanks for adding all of this @kenkehoe

PR #468, opened 2022-05-25
QC Configuration file

kenkehoe:
This PR is to complete the three ways a user can implement QC information to the data. This will apply the QC information stored in a YAML file by variable tests, to allow for user defined QC tests and ranges that are stored as time ranges instead of a test to apply. Also, added example how to use all the QC tools.


zssherman:
Hmm looks like some filepath issue for an example:
Extension error:
Could not find docstring in file "/home/runner/work/ACT/ACT/examples/plot_arm_qc.py". A docstring is required by sphinx-gallery unless the file is ignored by "ignore_pattern"
make: *** [Makefile:46: html] Error 2

kenkehoe:
@zssherman @AdamTheisen  I'm confused with this error. Do I need to do something else when adding an example file to the /examples directory?

zssherman:
@kenkehoe The error seems to be related now to:
aise InvalidURL(f"URL can't contain control characters. {url!r} "
http.client.InvalidURL: URL can't contain control characters. '/armlive/livedata/query?user=your username:your token&ds=sgpmfrsr7nchE11.b1&start=2021-03-29&end=2021-03-29&wt=json' (found at least ' ')

So it seems a path or string is wrong in the get arm data function retrieval

kenkehoe:
@zssherman  I'm confused with this. I can run the example code locally just fine. This last error appears to be an issue downloading the data. I wonder if the web-service was down when this ran. Is there a way to not actually run the example code? Do we need to run the code for an example?

AdamTheisen:
@kenkehoe taking a look now and rerunning it.  The code in the examples will run if it starts with "plot".  In looking at this, it could be an API error but it also could be with access to the secrets ARM_USERNAME/ARM_PASSWORD.  Let's see how this latest run works and go from there

AdamTheisen:
@kenkehoe so the "secrets" in GitHub are not available during pull requests due for security reason.  If you look at plot_ceil.py, you'll see how I've gotten around that but it does require a test file.  I swapped out the E38 MFRSR file for E11 and updated the unit test for it.  I also made some other updates.  Let's see how the tests do from that.

@mgrover1  @zssherman our tests area is up to 63MB right now which is included in our package on release.  Can we exclude the test directory from getting bundled into the release packages or is that needed?  I'm open to thoughts here.  My initial idea of using the web service to download data won't work for the reasons above.

zssherman:
@AdamTheisen The only reason it would be needed is for is if a user wanted to do unittests, but we can also exclude it and if they want them they can download them from the adc kinda like pyarts doc example files. I can't remember if conda unittests using act when testing the package, if it doesn't can always exclude.

AdamTheisen:
@kenkehoe Things are looking good!  My only thought is if we should rename the file from flag_qc.py, which is pretty generic and might give the wrong impression, to something like add_supplemental_qc.py.  Thoughts?

mgrover1:
@AdamTheisen we can look into excluding those from the packaging - I can go ahead and file a separate issue. Also, @kenkehoe I can review this

mgrover1 (approved):
I like this approach/implementation!

PR #473, opened 2022-06-08
ENH: Use lazy loading for loading in modules

### LINKED ISSUE #62: Speed up import time to load ACT
I think we need to pay more attention to the way we load ACT modules. Currently it takes about 8 seconds to load ACT. Also it would be better to not load modules that are not used. For example the cartopy module is loaded for any plot created even if it is not used. I think we should put that module import under the GeographicPlotDisplay class so it is only imported when that class is loaded. 

Not sure how this is typically done so if someone has suggestions I'm happy to implement.

mgrover1:
Currently, we **eagerly** import every module upon importing ACT (`discovery`, `qc`, etc.).

This means that when you call the

```python
import act
```

line, it will run through all of the ACT code to prepare to use. This can lead to **long** import times, on the order of ~10 seconds.

We can cut this down quite a bit by using the [SciPy SPEC 001](https://scientific-python.org/specs/spec-0001/), which outlines how to use the Lazy Loader package. 

The only caveat here is that
"Python 3.7, with [PEP 562](https://www.python.org/dev/peps/pep-0562/), introduces the ability to override module __getattr__ and __dir__. In combination, these features make it possible to again provide access to submodules, but without incurring performance penalties."

Which indicated this will **only work with Python 3.7+**

I included a first cut of this within the PR, and removed our 3.6 linux test since the only way we can merge this is by losing that test.

Closes #62


AdamTheisen:
@kenkehoe this might be what you've been looking for in terms of import speed/improvement

zssherman:
@mgrover1 Looks like coverage dropped too, maybe due to imports? colormap module etc

mgrover1:
> @mgrover1 Looks like coverage dropped too, maybe due to imports? colormap module etc

Yup - the coverage checks all the lines that are run. Since we are not running everything upon importing, we will need to add additional tests to cover everything (ex. colormap)

mgrover1:
@zssherman similar to the xarray module, we do need to eagerly load in the colormaps for matplotlib

mgrover1:
@zssherman these tests won't pass until it has the associated keys, which are only accessible when running on main (after the PR is merged)

mgrover1:
Probably - that makes sense!

AdamTheisen:
@mgrover1 @zssherman generally, the coverage has been about 1.5% less so there are still some areas where we are seeing a decrease in the coverage.  Let me look into it a little more on coveralls and see what's happening

act/discovery/__init__.py
```diff
@@ -4,8 +4,15 @@
...
-from .get_armfiles import download_data
-from .get_asos import get_asos
-from .get_cropscape import croptype
-from .get_airnow import get_airnow_bounded_obs, get_airnow_obs, get_airnow_forecast
-from .get_noaapsl import download_noaa_psl_data
+import lazy_loader as lazy
+
+__getattr__, __dir__, __all__ = lazy.attach(
+    __name__,
+    submodules=['get_armfiles', 'get_cropscape'],
```
AdamTheisen, line 11: Should we just have get_armfiles in here? Cropscape is not a high priority to load in initially

AdamTheisen:
Pinging @kenkehoe  here as well for awareness.

kenkehoe:
@AdamTheisen  I'm not really sure. How does this change the use of the read_netcdf() function use?

AdamTheisen:
@mgrover1 could you respond to Ken on this?

mgrover1:
> @AdamTheisen I'm not really sure. How does this change the use of the read_netcdf() function use?

Users can still use the same API

```python
import act

act.io.read_netcdf()
```

Is this what you were asking?

act/discovery/__init__.py
mgrover1: We should have all them here - this lazily loads in the discovery submodules (`get_armfiles`, `get_cropscape`, etc.)

AdamTheisen:
So, coveralls is saying there's a decrease in other modules but when I drill down into them there's no change.  I think the biggest change is that the following are not being loaded up at all initially here which is dropping the coverage.  As @mgrover1 noted, this should be resolved when the tests run on production.  I do think we should get an ENG in with ARM before we merge this to make sure there are no larger issues to think about.

Screenshot (3)

AdamTheisen (approved):

PR #484, opened 2022-06-23
Drop nan values with histogram display

### LINKED ISSUE #482: HistogramDisplay, plot_stairstep_graph encounters ValueError for range
I'm attempting to generate histograms for some Aerosol Optical Depth data. It is rather sparse, and looks like this:

[nan, nan, 0.04041248, nan, nan, nan, nan, nan, nan, 0.08403611, nan, nan, nan, nan, nan, 0.08443181, 0.05368716, nan, 0.07251927, 0.05726737, nan, 0.04878159, nan, 0.06544831, nan, nan, nan, nan, 0.04621224, nan, nan]

Plotting like so:
```
    var_name = 'aod_be_polyfit_415'
    nbins = 12
    try:
        ts_display = act.plotting.HistogramDisplay(
            nc_data,
            figsize=(19,5.5),
            subplot_shape=(1,),
        )
    except Exception as e:
        print(f"{type(e)} error generating histogram of netCDF data for {var_name}:\n{e}")
    else:
        ts_display.plot_stairstep_graph(var_name, bins=nbins, subplot_index=(0,))
```

It encounters this error:
Traceback (most recent call last):
  File "plot_aodbe_daily_avgs.py", line 335, in <module>
    sys.exit(main())
  File "plot_aodbe_daily_avgs.py", line 320, in main
    ts_display.plot_stairstep_graph(var_name, bins=nbins, subplot_index=(0,))
  File "/home/<username>/.local/lib/python3.9/site-packages/act/plotting/histogramdisplay.py", line 386, in plot_stairstep_graph
    my_hist, bins = np.histogram(
  File "<__array_function__ internals>", line 180, in histogram
  File "/home/<username>/.local/lib/python3.9/site-packages/numpy/lib/histograms.py", line 793, in histogram
    bin_edges, uniform_bins = _get_bin_edges(a, bins, range, weights)
  File "/home/<username>/.local/lib/python3.9/site-packages/numpy/lib/histograms.py", line 426, in _get_bin_edges
    first_edge, last_edge = _get_outer_edges(a, range)
  File "/home/<username>/.local/lib/python3.9/site-packages/numpy/lib/histograms.py", line 323, in _get_outer_edges
    raise ValueError(
ValueError: autodetected range of [nan, nan] is not finite

Since np.histogram accepts a `range` parameter, I assume this error is because it attempts to auto-calculate the range, but the data contains nans and throws that off. I experimented and found if I replace nan with e.g. 1.0, the plotting works. I have to replace all nans.

I imagine there might be various ways to address this.
* Can this be corrected within ACT, so either A) ACT determines the range better than numpy does, ignoring the nans, or B) accepts a range parameter that it will pass to np.histogram?
* Is there a way you might recommend handling this in the data before attempting to plot a histogram? As I noted, I could replace all nans with other values (-9999), but then there'd need to be a way to mask out that value... and might as well just mask the nans.

mgrover1:
Closes #482

### DIFF
--- a/act/plotting/histogramdisplay.py
+++ b/act/plotting/histogramdisplay.py
@@ -83,6 +83,11 @@ class HistogramDisplay(Display):
         self.axes[subplot_index].set_ylim(yrng)
         self.yrng[subplot_index, :] = yrng
 
+    def _get_data(self, dsname, fields):
+        if isinstance(fields, str):
+            fields = [fields]
+        return self._obj[dsname][fields].dropna('time')
+
     def plot_stacked_bar_graph(
         self,
         field,
@@ -139,15 +144,16 @@ class HistogramDisplay(Display):
         elif dsname is None:
             dsname = list(self._obj.keys())[0]
 
-        xdata = self._obj[dsname][field]
+        if sortby_field is not None:
+            ds = self._get_data(dsname, [field, sortby_field])
+            xdata, ydata = ds[field], ds[sortby_field]
+        else:
+            xdata = self._get_data(dsname, field)
 
         if 'units' in xdata.attrs:
             xtitle = ''.join(['(', xdata.attrs['units'], ')'])
         else:
             xtitle = field
-
-        if sortby_field is not None:
-            ydata = self._obj[dsname][sortby_field]
 
         if bins is not None and sortby_bins is None and sortby_field is not None:
             # We will defaut the y direction to have the same # of bins as x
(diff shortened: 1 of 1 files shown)


zssherman (approved):
Looks good to me!

AdamTheisen:
@mgrover1 for the test_histogram_errors failure we can just add in a time variable to the dummy dataset we are using.  For the test_size_distribution plot, the image is different so we will need to figure out what the difference is and if it's expected.  Thoughts?

zssherman:
@AdamTheisen A comparison of the two plots, if they new one is expected or enhanced behavior, we can switch out the test image, we had to recently do that for one of Ken's PRs

AdamTheisen:
All looks good now, going to merge in.

PR #490, opened 2022-07-07
ADD: Add new PSL Wind Profiler Temperature Reader

### LINKED ISSUE #489: Make the NOAA PSL reader more generic
The current reader only works with wind profiler data from the CTD site (see https://github.com/ARM-DOE/ACT/blob/main/act/io/noaapsl.py#L36)... we need to make this more generic to support other sites, and other date types from this site since the files are similar in structure and the name of the reader is "NOAA PSL" which covers a wide range of data

mgrover1:
Add a new reader (closes #489 )


zssherman (approved):

mgrover1:
Let's wait for the tests to see if things are working!

mgrover1:
I'd also like to add an example in here...

mgrover1:
@zssherman - on second thought, let's delay creating the example.. I think it would be better to have one once we have a better idea of what needs refactoring/such

PR #494, opened 2022-07-21
ENH: Support searching at the second level for ARM data

### LINKED ISSUE #493: ENH: Enable Hour/Minute Level of Searching with ARMLive API
Currently, the hour/minute/second field of the ARM data access API is not being passed to the query, preventing the ability to search at this level. This is an issue when working with larger datasets (ex. x-band radar) where a day of data is 10s of GB in size and a user wants a small subset.

I put together some new logic over in this PR (https://github.com/pangeo-forge/staged-recipes/pull/153) for the Pangeo Forge recipe.

This would be good to bring into ACT + update so users can use more refined searches.

mgrover1:
Closes #493

### DIFF
--- a/act/discovery/get_armfiles.py
+++ b/act/discovery/get_armfiles.py
@@ -31,10 +31,14 @@ def download_data(username, token, datastream, startdate, enddate, time=None, ou
         The name of the datastream to acquire.
     startdate : str
         The start date of the data to acquire. Formats accepted are
-        YYYY-MM-DD, DD.MM.YYYY, DD/MM/YYYY, YYYYMMDD or YYYY/MM/DD.
+        YYYY-MM-DD, DD.MM.YYYY, DD/MM/YYYY, YYYYMMDD, YYYY/MM/DD or
+        any of the previous formats with THH:MM:SS added onto the end
+        (ex. 2020-09-15T12:00:00).
     enddate : str
         The end date of the data to acquire. Formats accepted are
-        YYYY-MM-DD, DD.MM.YYYY, DD/MM/YYYY, YYYYMMDD or YYYY/MM/DD.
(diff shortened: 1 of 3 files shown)


zssherman (approved):
Looks good to me!

act/discovery/get_armfiles.py
```diff
@@ -31,10 +31,14 @@ def download_data(username, token, datastream, startdate, enddate, time=None, ou
...
+        YYYY-MM-DD, DD.MM.YYYY, DD/MM/YYYY, YYYYMMDD, YYYY/MM/DD or
+        any of the previous formats with THH:MM:SS added onto the end
+        (ex. 2020-09-15T12:00:00).
     enddate : str
         The end date of the data to acquire. Formats accepted are
-        YYYY-MM-DD, DD.MM.YYYY, DD/MM/YYYY, YYYYMMDD or YYYY/MM/DD.
+        YYYY-MM-DD, DD.MM.YYYY, DD/MM/YYYY, YYYYMMDD or YYYY/MM/DD, or
+        any of the previous formats with THH:MM:SS added onto the end
+        (ex. 2020-09-15T13:00:00).
     time: str or None
```
AdamTheisen, line 42: Do we still want to keep this for legacy purposes or could we remove this and just use the start/enddates above to download one file?

AdamTheisen (changes requested):
@mgrover1 should we add some unit tests in to cover this addition?  Other than that, looks good!

act/discovery/get_armfiles.py
mgrover1: I think we can leave it in for legacy purposes... otherwise people's workflows may break

mgrover1:
Thanks @AdamTheisen - I added a unit test, waiting to see if everything passes

AdamTheisen (approved):

PR #495, opened 2022-07-21
FIX: Add more robust parsing method to PSL Temperature data

### LINKED ISSUE #491: read_psl_wind_profiler_temperature not reading full file
It looks like the new function to read NOAA PSL temperature is not reading the full file.  For instance, the file linked below has 2 times in it 172536 and 175538 UTC but only the first one is being read in.  We need to update it to read in all times in the file.

https://downloads.psl.noaa.gov/psd2/data/realtime/Radar449/WwTempSubHourly/ctd/2022/196/ctd22196.17t

mgrover1:
Closes #491

### DIFF
--- a/act/discovery/get_armfiles.py
+++ b/act/discovery/get_armfiles.py
@@ -96,10 +96,10 @@ def download_data(username, token, datastream, startdate, enddate, time=None, ou
     # start and end strings for query_url are constructed
     # if the arguments were provided
     if startdate:
-        start = date_parser(startdate, output_format='%Y-%m-%dT%H:%M:%SZ')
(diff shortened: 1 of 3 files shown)


AdamTheisen (approved):
@mgrover1 this works well!  The only thing I would leave to you is if we need to adjust the coordinates so that it plays nice with xarray as well.  Right now a obj['Tc'].plot() puts time on the y-axis.

image

mgrover1:
@AdamTheisen - the fix here is to reorder the data so height is firs, then time. I add the `.transpose()` line which fixes this in the recent commit

PR #496, opened 2022-07-25
FIX: Fix hour minute second format for ARMLive

mgrover1:
For the ARMLive API, we are missing colons in between hour, minute, and second


zssherman (approved):

mgrover1:
Adding one more test here for this...

PR #497, opened 2022-07-25
FIX: Fix failing ci by adding missing % to ARMLive

mgrover1:
Fix the failing CI caused by a missing % sign in the strftime call in the ARMLive url building


zssherman (approved):
Looks good to me!

PR #498, opened 2022-07-25
FIX: Pass datetime to function not str

mgrover1:
It looks like it was passing the datetime string instead of the datetime object... this fixes that


AdamTheisen (approved):
Update looks good!

zssherman (approved):

PR #508, opened 2022-08-15
Adding new plot functionality to Data Rose

AdamTheisen:
New plot will display the mean of the variable based on the wind direction and speed in a polar contour plot.  Updated the tests and the example to showcase this feature.


AdamTheisen:
For example, looking at the [baseline plot](https://github.com/ARM-DOE/ACT/blob/1c08edb99a517b2359fb4c67f6925c6f6db9bb98/act/tests/baseline/test_plot_datarose.png?raw=true), you can see that the higher temperatures are when the winds are out of the South which is to be expected in Oklahoma.

act/plotting/windrosedisplay.py
```diff
@@ -302,6 +304,14 @@ def plot_data(
             but other options are 'Median' and 'Stdev'
         clevels : int
             Number of contour levels to plot
+        contour_type : str
```
mgrover1, line 307: Can we make this `aggregation` to be consistent with pandas/xarray APIs? https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.aggregate.html

act/plotting/windrosedisplay.py
```diff
@@ -374,18 +384,68 @@ def plot_data(
...
-            )
-            hist = np.insert(hist, -1, hist[0], axis=0)
-            cplot = self.axes[subplot_index].contourf(
-                np.deg2rad(xedges), yedges[0:-1], np.transpose(hist),
-                cmap='rainbow', levels=clevels, **kwargs
-            )
-            cbar = self.fig.colorbar(cplot, ax=self.axes[subplot_index])
-            cbar.ax.set_ylabel('Count')
-            plot_type_str = 'Heatmap of'
+            if contour_type == 'Count':
```
mgrover1, line 387: Also - these should be lowercase

mgrover1 (changes requested):
One small suggestion - I think we should be consistent with the xarray and pandas APIs when it comes to aggregation.

act/plotting/windrosedisplay.py
AdamTheisen: @mgrover1 I'm not sure I'm following on this request.  Can you explain a little what you are looking for?

mgrover1: Within xarray, pandas, and the broader ecosystem, the term usually used for to specify the type of binning/data reduction is `aggregation`. I am suggesting we use similar terminology here, to be consistent with other APIs such as https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.aggregate.html and https://foundations.projectpythia.org/core/xarray/computation-masking.html#aggregation-methods

AdamTheisen: Okay, so you are just recommending that we rename that keyword instead of reworking the plotting protocols to follow something like this `display.plot_datarose().mean()`, right?

mgrover1: On second thought, I think it is fine as is...! Sorry for the confusion...

mgrover1 (approved):

PR #512, opened 2022-08-21
ENH: Adding keyword to csv read to help with some other cases

AdamTheisen:
Adding ignore_index keyword for pd.concat to handle more use cases


zssherman:
Looks good to me @AdamTheisen!

PR #515, opened 2022-08-22
ENH: Updating example to try and get it to run

AdamTheisen:
Example is not working correctly


zssherman (approved):

zssherman:
Looks good to me, was the example not generating for docs? Or was there an error?

AdamTheisen:
@zssherman sorry, I missed your question! I think it was due to the issue with using the token to get data from the API

PR #519, opened 2022-08-26
BUG: Bug fix for the FMCW reader that was adding an extra day to the …

AdamTheisen:
…time


zssherman (approved):

AdamTheisen:
@zssherman would we want to do a micro release on this bug since it could impact the science?

zssherman:
@AdamTheisen No worries, I can do that!

PR #520, opened 2022-08-30
DOC: Adding example for both fmcw and parisvel for august event.

zssherman:


AdamTheisen:
@zssherman my only thought is if we should replace this example plot_noaa_fmcw_moment.py  with the one here.  We could just delete as it is duplicate.  Thoughts?

zssherman:
@AdamTheisen Yeah we can do that, give me one moment and i'll switch it out

AdamTheisen (approved):

Issue #487, opened 2022-06-29
Continue readers for NOAA PSL Data


AdamTheisen:
There's a lot of different instruments deployed in the NOAA PSL Netowrk, developing readers for these would be beneficial, including with the upcoming AMF3 campaign coming up that will be in the vicinity of Courtland Airport

https://psl.noaa.gov/data/obs/sites/view_site_details.php?siteID=ctd

AdamTheisen:
Checklist of readers that need to be developed
- [ ] 915 MHz Wind Profiler Wind
- [ ] 915 MHz Wind Profiler Temperature
- [ ] 449 MHz Wind Profiler Wind
- [ ] 449 MHz Wind Profiler Temperature
- [ ] 3 GHz S-Band Moments
- [ ] Disdrometer Parsivel Optical
- [ ] Surface Meteorology (Data Logger CR100 link) - This has most of the other variables so would reduce readers
- [ ] GPSTrimble NetRS (Water Vapor)

mgrover1:
With #490 merged, we now have two readers in! Planning on taking a look at the other frequency wind profilers after SciPy to make sure the current readers support those...

kenkehoe:
I've been looking at SurfaceMet for NOAA PSL data as part of SAIL. I have an idea for the generic reader. Turns out the downloader pulls from the RAW area which does not have file headers in the file or documentation on the data. I'm looking into understanding this now and have an idea on how to resolve the different files.

AdamTheisen:
The FMCW read might work for the S-band moments as well.  Will have to test a little though

kenkehoe:
@AdamTheisen I think my recent PR resolves Surface Meteorology (Data Logger CR100 link). Shall we check that one off the list?

AdamTheisen:
Done!

AdamTheisen:
I'm removing the GPS one from the list for now.  The documentation does not line up with the readable file so there is no way of telling that we are parsing it correctly.  Once #559 is merged in we can check the last one off the list and close out.

zssherman:
Finished, closing

PR #524, opened 2022-09-09
Psl met

kenkehoe:
This is to integrate a configuration file for the NOAA PSL met data. Most of the used data is contained in the new configuration file, but there are many locations which are not in the configuration file. We can add more as needed/requested as the additional sites are easy to add. A user can make a copy of the configuration file, edit and use the keyword to point to the updated version if they need to read a file and can not wait for the file to be updated.


kenkehoe:
PR for #487

act/io/noaapsl.py
```diff
@@ -344,6 +348,131 @@ def parse_date_line(list_of_strings):
...
+    Parameters
+    ----------
+    filenames : str, list of str
+        Name of file(s) to read.
+    conf_file : str or Pathlib.path
+        Default to ./conf/noaapsl_SurfaceMet.yaml
+        Filename containing relative or full path to configuration
+        YAML file used to describe the file format for each PSL site.
+        If the site is not defined in the default file, the default file
+        can be coppied to a local location, and the missing site added.
```
zssherman, line 365: coppied should be copied

zssherman (approved):
Besides the one typo I see, everything looks good to me.

Issue #522, opened 2022-09-07
Expand add_nan keyword in plot() call for line  plots.


kenkehoe:
Currently the add_nan keyword will fill in periods where data is missing larger than a time step to ensure the 2D plot does not stretch the colors across the missing data period. This will insert a column of NaN values to stop connecting the data points. The default is to plot lines with markers only, no lines. But if a linestyle is chosen and there is a large data gap, the plot will appear to have a straight line connecting the data on either side of the missing data period. We should expand the existing keyword add_nan to insert a single NaN for line plotting as well.

kenkehoe:
Updated with PR #525

PR #525, opened 2022-09-09
Add in nan

kenkehoe:
#522 

Adding in function to existing add_nan keyword in plot() function to find missing data periods in 1D data and add a NaN to stop plotting a connecting line between two data points that should not have a line connecting.

Also, discovered an issue with the add_in_nan() function. It was not correctly inserting a NaN for instances after first found data gap period.


act/tests/test_utils.py
```diff
@@ -28,7 +28,12 @@ def test_dates_between():
 
 
 def test_add_in_nan():
-    # Make a 1D array with a 4 day gap in the data
+
+    print()
+
+
```
zssherman, line 34: I would remove this whitespace, and is the print needed?

zssherman (approved):
Added a comment, but besides that looks good.

AdamTheisen:
Looks good, going to merge in

PR #528, opened 2022-09-16
ENH: Updating to remove one attribute and set another to a string so …

AdamTheisen:
…that it can appropriately write out to netcdf


jrobrien91 (changes requested):
Should 'Dependent_Var_Num' also be set to a string? Otherwise this looks good.

AdamTheisen:
@jrobrien91 it's an int right now so I think that's fine unless you want it to be something other than a length.  I'm not quite sure what it's supposed to be, number of dependent variables?

PR #534, opened 2022-09-28
ENH: Bug fix for inverted_y_axis

AdamTheisen:
Found an issue that the invert_y_axis keyword was not working correctly for a 2 panel plot.


zssherman (approved):

zssherman:
@AdamTheisen PR looks good, error looks to be a datetime issue, but only for one tests, let me restart it.

AdamTheisen:
Updating the tests to do an inverted plot so we can catch this in the future if something happens.

AdamTheisen:
Approved, merging in.

PR #537, opened 2022-10-07
DOC: Add blog post for comparing ARM and NOAA data for the SAIL campaign.

zssherman:


mgrover1 (changes requested):
Great content! Just a few suggestions... I would recommend using this one as an example to base the formatting/sections on https://arm-doe.github.io/pyart/blog_posts/2022/ka-band-pyart-xarray-hvplot.html

zssherman:
@AdamTheisen On the science side of things, what would be a good motivation for these examples for the images we used in the ACT for science comparing the noaa and arm instruments? Per @mgrover1 suggestions above.

AdamTheisen:
@zssherman I would say that the biggest benefit is not necessarily the comparison but the ability to easily bring together data from different organizations all together in one place for exploring and analyzing.  This expands the information available to scientists in their research and also allows for quality control checks on the data by comparing it which increases the confidence in the quality and calibration of the instrumentation.

zssherman:
Cool thanks @AdamTheisen, I added that in as well as the changes suggested by @mgrover1

mgrover1 (approved):
Thanks for making those changes - looks great!

PR #540, opened 2022-10-11
MNT: Update to check for a level files etc.

### LINKED ISSUE #532: ARM Data API Returns Text when File not Available
I tried to download gucceilpblhtM1.a0 but it's not available through the webservice due to it being an a0-level file.  Instead the API returned text that was included in the file.  We should put in a check to ensure that these cases are caught and files are not produced.

```
This data file is not available on /data/archive.  To download this file, please an order via Data Discovery. https://adc.arm.gov/discovery
```

zssherman:
Closes #532 , @AdamTheisen This now checks for that message and then either raises a download error or continues on and downloads the files. I'm not sure if this is the most robust way.

### DIFF
--- a/act/discovery/get_armfiles.py
+++ b/act/discovery/get_armfiles.py
@@ -4,6 +4,7 @@ Script for downloading data from ARM's Live Data Webservice
 """
 
 import argparse
+import contextlib
 import json
 import os
 import sys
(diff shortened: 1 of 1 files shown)


act/discovery/get_armfiles.py
```diff
@@ -143,22 +143,35 @@ def download_data(username, token, datastream, startdate, enddate, time=None, ou
...
                 'https://adc.arm.gov/armlive/livedata/' + 'saveData?user={0}&file={1}'
             ).format(':'.join([username, token]), fname)
             output_file = os.path.join(output_dir, fname)
-            # make directory if it doesn't exist
-            if not os.path.isdir(output_dir):
-                os.makedirs(output_dir)
             # create file and write bytes to file
             with open(output_file, 'wb') as open_bytes_file:
-                open_bytes_file.write(urlopen(save_data_url).read())
+                if 'this data file is not available' in urlopen(
```
AdamTheisen, line 153: @zssherman  should we check this before opening the output file for writing?

AdamTheisen:
@zssherman there are some Python 3.6 errors showing up so we might have to wait to merge this.

ERROR act/tests/test_correct.py - NameError: name 'DistributionNotFound' is n...

AdamTheisen (approved):

PR #553, opened 2022-11-03
FIX: Fix issues with lazy loading accessors

### LINKED ISSUE #552: Accessor not available in dataset
I fetched the latest updates after the lazy_loading PR and ran pytest and am seeing a lot of errors with accessors not loading.  Clean, QCFilter, and QCTests are no longer available in the datasets for some reason.

FAILED test_io.py::test_io_mfdataset - AttributeError: 'Dataset' object has no attribute 'clean'
FAILED test_io.py::test_io_write - AttributeError: 'Dataset' object has no attribute 'clean'
FAILED test_io.py::test_clean_cf_qc - AttributeError: 'Dataset' object has no attribute 'clean'

mgrover1:
Closes #552 

@AdamTheisen accessors need to eagerly loaded in... there is not a performance decrease to doing this which is nice. This fixes the issue.

### DIFF
--- a/act/__init__.py
+++ b/act/__init__.py
@@ -5,11 +5,13 @@ ACT: The Atmospheric Community Toolkit
 """
 
 import lazy_loader as lazy
+
 # No more pandas warnings
 from pandas.plotting import register_matplotlib_converters
 
 from . import tests
 from ._version import get_versions
+from .qc import QCFilter, QCTests, clean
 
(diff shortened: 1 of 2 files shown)


AdamTheisen (approved):
Code coverage decreased more than normal but coveralls wasn't showing me anything new so I bumped up the threshold and am rerunning the tests

PR #556, opened 2022-11-04
Bug: Fix for when files are not available through web service.

AdamTheisen:
Fix for #532  when files are not available through the web service.


mgrover1 (approved):
This approach looks great!

PR #559, opened 2022-11-11
ENH: Adding reader for S-band and refactored code so the sband and fm…

AdamTheisen:
…cw reader use the same code base.


zssherman (approved):

zssherman:
Looks good to me merging

PR #562, opened 2022-11-15
MNT: Add directories for downloaded test files locally.

zssherman:


zssherman:
@AdamTheisen am understanding these datasets correctly that they are downloaded while running the unitteste?

AdamTheisen:
@zssherman Yes!  That is correct with these NOAA files.  Just downloaded for the testing.  We probably need a cleanup function for these to be run after the tests.

PR #565, opened 2022-11-15
FIX: Figure size wasn't being set correctly in Xsectiondisplay

zssherman:
subplot was being overwritten and figure size would not change the plot when calling figsize in XsectionDisplay. This has been fixed as well as the example.


zssherman:
Trying to figure out how to replace the baselines...

AdamTheisen:
@zssherman if you run this, it will recreate all the baseline images. Then you can just check the cross-section plots and commit them.  

```
pytest --mpl-generate-path=baseline
```

zssherman:
@AdamTheisen Ah gotcha, thanks! I will do that!

mgrover1 (approved):
Good move setting the figsize

PR #567, opened 2022-11-16
FIX: Updates setuptools install to fix warning.

zssherman:
setuptools now recommends find_namespace_packages for finding modules in the setup.py


mgrover1 (approved):
Great! Good catch.

zssherman:
I updated the environment and requirements as well to include lazy loader and removed unused lxml

PR #568, opened 2022-11-16
TST: Updating test suite for pytest skip if for metpy.

zssherman:
Tests failed with metpy not installed, which shouldn't be the case, fixing functions to follow pytest skipif


AdamTheisen (approved):
I like this!  We should look at doing this for the ones that need the Github Secrets as well which is mostly for pull ARM and AIRNOW files.

zssherman:
@AdamTheisen Yeah that's a great idea!

PR #572, opened 2022-11-17
FIX: lxml was needed and scikit and metpy qc tests weren't optional.

zssherman:
lxml is needed despite not existing in the ACT codebase itself but used by pandas. And fixing tests that weren't optional.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,3 +17,4 @@ lazy_loader
 fsspec
 lazy_loader
 icartt
+lxml
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,5 +16,4 @@ netcdf4
 lazy_loader
 fsspec
 lazy_loader
-icartt
 lxml
--- a/.github/workflows/python-package-conda.yml
+++ b/.github/workflows/python-package-conda.yml
@@ -51,10 +51,6 @@ jobs:
           use-mamba: true
           miniforge-variant: Mambaforge
 
-      - name: Install ICARTT
-        run:
-          python -m pip install icartt
-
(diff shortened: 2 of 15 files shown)


AdamTheisen:
@zssherman  I see that icartt was added to the requirements file.  I think we were planning icartt as an optional dependency right?  Did we miss that with the icartt PR?

zssherman:
@AdamTheisen Yep! You are right, a mistake on my part, ill remove that from the requirements

requirements.txt
```diff
@@ -16,4 +16,5 @@ netcdf4
 lazy_loader
 fsspec
 lazy_loader
-icartt
\ No newline at end of file
+icartt
```
mgrover1, line 19:
```suggestion
```

AdamTheisen (approved):

zssherman:
@AdamTheisen @mgrover1 This should now be updated to have metpy as a dependency, readme are updated to have icartt pyart and scikitposthocs optionals etc.

AdamTheisen:
@zssherman  I think this PR is good to go but I'm going to open a new issue as we have Py-ART in the feedstock but list it as an optional dependency.  Our test in test_utils.py does not check for pyart install.

zssherman:
I can add a check for the test and remove it from the feedstock @AdamTheisen if that's what we want to do.

mgrover1 (approved):
Looks good to me

AdamTheisen:
@zssherman I think that's what the intent originally was even though I didn't quite implement it that way from the looks of it.

PR #579, opened 2022-11-20
Adding functions to the __init__.py files

AdamTheisen:
We were missing some of the functions in the __init__.py files.  We might want to do an inventory to make sure we aren't missing anything else.

### DIFF
--- a/act/qc/__init__.py
+++ b/act/qc/__init__.py
@@ -8,8 +8,6 @@ import lazy_loader as lazy
 
 # We need to import clean first to register the accessor
 from .clean import *
-from .qcfilter import QCFilter
-from .qctests import QCTests
 
 __getattr__, __dir__, __all__ = lazy.attach(
     __name__,
@@ -19,12 +17,21 @@ __getattr__, __dir__, __all__ = lazy.attach(
         'bsrn_tests',
         'comparison_tests',
         'qcfilter',
+        'qctests',
         'radiometer_tests',
         'sp2',
     ],
     submod_attrs={
         'arm': ['add_dqr_to_qc'],
-        'qcfilter': ['set_bit', 'unset_bit', 'parse_bit'],
+        'qcfilter': [
+            'QCFilter',
+            'parse_bit',
+            'set_bit',
+            'unset_bit',
+        ],
+        'qctests': [
+            'QCTests',
+        ],
         'radiometer_tests': [
             'fft_shading_test',
             'fft_shading_test_process',
--- a/act/qc/__init__.py
+++ b/act/qc/__init__.py
@@ -25,9 +25,11 @@ __getattr__, __dir__, __all__ = lazy.attach(
     ],
     submod_attrs={
         'arm': ['add_dqr_to_qc'],
-        'qcfilter': ['QCFilter'],
-        'qctests': ['QCTests'],
-        'radiometer_tests': ['fft_shading_test'],
+        'qcfilter': ['set_bit', 'unset_bit', 'parse_bit'],
+        'radiometer_tests': [
+            'fft_shading_test',
+            'fft_shading_test_process',
+        ],
         'bsrn_tests': ['QCTests'],
         'comparison_tests': ['QCTests'],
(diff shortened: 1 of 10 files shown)


act/qc/__init__.py
```diff
@@ -25,9 +25,11 @@
     ],
     submod_attrs={
         'arm': ['add_dqr_to_qc'],
-        'qcfilter': ['QCFilter'],
```
zssherman, line 28: do we still need to specify QCFilter and QCtests

AdamTheisen: @zssherman  good question.  I took out QCTests from the modules but we need to keep QCFilter in there because there are 3 functions outside the class that we need to load up that we are not currently doing so.

AdamTheisen:
I am finding a few more issues with the documentation so working through things.  Hold on any approvals/merging

AdamTheisen:
@zssherman  I think I have all the issues solved.  @mgrover1 would you want to take a look as well?

mgrover1 (approved):
This looks good - I can put together a list of all of the modules and associated submodules to make sure we don't miss anything.

AdamTheisen:
Hold for one sec.  When removing the QCFilter, it no longer is included in the documentation.  Let me look into this

act/qc/__init__.py
```diff
@@ -19,15 +19,16 @@
         'bsrn_tests',
         'comparison_tests',
         'qcfilter',
-        'qctests',
```
mgrover1, line 22: Why did you remove qctests here?

act/qc/__init__.py
```diff
@@ -19,15 +19,16 @@
...
         'radiometer_tests',
         'sp2',
     ],
     submod_attrs={
         'arm': ['add_dqr_to_qc'],
-        'qcfilter': ['QCFilter'],
-        'qctests': ['QCTests'],
```
mgrover1, line 29: same here

act/qc/__init__.py
AdamTheisen: @mgrover1 we are doing direct imports of QCTests so it should be available.  Do we need to keep it in the lazy_loader section if we are already importing it?

```
from .qcfilter import QCFilter
from .qctests import QCTests
```

act/qc/__init__.py
AdamTheisen: @mgrover1 see previous response.  If we are importing it directly at the start of the __init__ do we need it in the lazy_loader modules?

act/qc/__init__.py
mgrover1: Good point - it looks like it is showing up in the built documentation in the github action log

mgrover1:
@AdamTheisen does your main __init__.py file in `act/__init__,py` include this?

https://github.com/ARM-DOE/ACT/blob/main/act/__init__.py#L14

AdamTheisen:
@mgrover1 yes it does.  Looks like we have it in both the act level and qc level.

https://github.com/ARM-DOE/ACT/blob/829c5a1fffcd5b690ccbe493fbdcc62b52b8bd10/act/qc/__init__.py#L11

mgrover1:
@AdamTheisen and the docs aren't showing this on your local act doc build?

AdamTheisen:
@mgrover1 I'm was just going off of what was already in there.  I do see some weirdness though in the QCFilter documentation in that is has all the QCTest functions as well. https://arm-doe.github.io/ACT/search.html?q=QCFilter

mgrover1:
@AdamTheisen is this building locally

> @AdamTheisen and the docs aren't showing this on your local act doc build?

AdamTheisen:
@mgrover1 this is building locally now.  I think I got all the issues resolved that were impacting my builds.

mgrover1:
Looks good to me!

zssherman:
@AdamTheisen Good to go then? I can merge this and get a release out.

zssherman (approved):

AdamTheisen:
@zssherman yep, I think we're good to go here

Issue #584, opened 2022-12-01
Useful feature: Load ARM dod from file on computer


rcjackson:
It would be useful to have the create_arm_obj_from_dod function be able to load a DOD file from the computer instead of from PCM. Sometimes PCM is not accessible, like from a compute node on cumulus. Therefore having ACT be able to use a downloaded dod file would be great here. I can work on that feature.

mgrover1:
Closed with #589

PR #589, opened 2022-12-06
ADD: Ability to load DOD from file for create_obj_from_arm_dod

rcjackson:
This resolves issue #584 where the user can now specify a file to load the DOD from when creating an ACT object from a DOD. The DOD filename is specified in the proc string, and if the local_file flag is set to True then ACT will load the DOD from the file location instead of from PCM.


zssherman:
Thanks @rcjackson for putting this together! Looks like it's failing to pep8 stuff

zssherman (approved):
Approved, is the drop due to this PR?

zssherman:
@AdamTheisen is this what happened to that other PR? Its showing a drop in get_airnow

AdamTheisen:
@zssherman that is strange.  It shouldn't fluctuate that much from PR to PR as it's been consistently around 1.7 now.  I'm rerunning again and will see what happens.  Coveralls is not really showing any differences besides the known ones.

zssherman:
Yeah, the code itself is sound, once the tests pass we can merge and see if the coverage resolves

AdamTheisen:
@zssherman  I would say that once these tests all pass we force the merge and see what happens on production.  I can always update the threshold again for coveralls

PR #595, opened 2022-12-16
Bug fix for the compute_winds_from_ppi code

AdamTheisen:
There was a bug in the code that set the snr_name when there were processes dependent on the snr_name.  So when multiple datasets were in an object, the first one ran correctly but the rest did not.  This should resolve that.  I also upgraded it to use dask processing to speed it up.  One day took about 4 seconds to process which was a significant improvement from before which took 10+ seconds.

### DIFF
--- a/act/retrievals/doppler_lidar.py
+++ b/act/retrievals/doppler_lidar.py
@@ -3,7 +3,7 @@ Functions for doppler lidar specific retrievals
 
 """
 import warnings
-
+import dask
 import numpy as np
 import xarray as xr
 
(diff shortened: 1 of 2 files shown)


AdamTheisen:
@kenkehoe this should not impact existing codes in how you are calling it at the DQ Office but wanted you to be aware.  Hopefully you see a speed boost!

kenkehoe:
Awesome, thanks for fixing this.

zssherman (approved):

zssherman:
@AdamTheisen Looks good to me, thanks for putting this together!

PR #598, opened 2022-12-20
DOC: Reorganize examples

### LINKED ISSUE #566: Improve navigating the examples
Currently the examples README.txt  file has a link to the Discussions area. I think the formatting is wrong as the hyperlink returns a 404 error. But I think we should update the README.txt file to more of the heavy lifting to point the user to the correct example than pointing the user to the Discussion area. We may need to rename some of the examples as everything starts with plot_* but not every example generates a plot.

Also, recent PR #563 created a new example to show how to downsample the data before plotting to speed up 2D plotting. I would not know which example to look in to find this information from the name of the file, plot_examples.py. We should create a TOC or some other listing of the problem/example and hyperlink to that example in the examples/README.txt

mgrover1:
Closes #566


mgrover1:
@AdamTheisen - @zssherman and I sat down and reorganized the examples... let me know what you think!

zssherman (approved):

AdamTheisen (commented):
I know the plot_qc_example.py example shows QC data but it's really intended to show the visualization of QC information so I think it would be better suited under plotting but open to feedback here.

examples/qc/plot_state_variable.py
```diff
@@ -9,10 +9,11 @@
 
 """
 
+from matplotlib import pyplot as plt
```
AdamTheisen, line 12: I think this one should be in plotting as well

examples/retrievals/plot_skewt.py
```diff
@@ -8,9 +8,9 @@
 """
```
AdamTheisen, line 10: Can we have this in plotting as well?  We do some retrievals here but not sure it's the focus of it.

AdamTheisen:
@mgrover1 @zssherman looks good!  I like the addition of a workflows area!  I just had a couple thoughts on where some of these might be better suited but open to thoughts/feedback.

mgrover1:
Thanks for the suggestions @AdamTheisen ! I made those changes.

AdamTheisen (approved):

mgrover1:
Once the tests pass, I will merge @AdamTheisen @zssherman

Issue #604, opened 2023-01-04
Colorblind friendly QC Defaults


AdamTheisen:
The QC plot generation defaults to red/green for incorrect/correct data.  We should really come up with a keyword to pass in to plot a colorblind friendly version or change the default to be colorblind friendly.

zssherman:
I second this, I would say the default, but maybe @kenkehoe can have insight for the data quality folks.

kenkehoe:
I'm happy to update the colors. I think leaving red, yellow, green as default is best since ARM has a long history with those colors. What colors do you suggest?

AdamTheisen:
@kenkehoe I have no experience here but could shifting the green for good to a light blue be an option?  @mgrover1 @zssherman thoughts here?

zssherman:
@AdamTheisen Could try switching out the green red for their equivalents in homeyer or chase spectral as an argument, if its best for maintain the ARM standards as Ken mention.

AdamTheisen:
@zssherman  @kenkehoe  I took some of the RGB values from the Homeyer CMAP function and tested that out. What are your thoughts on this as a cb_friendly keyword.

<img width="1314" alt="image" src="https://user-images.githubusercontent.com/13421074/215193316-d9f4834d-3c0a-4a54-94ea-4a5322d67794.png">

zssherman:
@AdamTheisen Beat me to it! I think having it as an argument would be good for now. But yeah I think this is a good start!

AdamTheisen:
The user can always change the colors themselves with the existing options but this might be a quick method for them to change it.

kenkehoe:
Works for me.

AdamTheisen:
Closing with #613

PR #613, opened 2023-01-27
Cb friendly

AdamTheisen:
This would close out #604  but introduces a keyword for the timeseries plot and the qc plot to use colorblind friendly colors.  RGB values were taken from the HomeyerRainbow cmap.

### DIFF
--- a/act/plotting/timeseriesdisplay.py
+++ b/act/plotting/timeseriesdisplay.py
@@ -288,6 +288,7 @@ class TimeSeriesDisplay(Display):
         secondary_y=False,
         y_axis_flag_meanings=False,
         colorbar_labels=None,
+        cb_friendly=False,
         **kwargs,
     ):
         """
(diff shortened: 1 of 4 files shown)


AdamTheisen:
Examples plot changes

ceil_homeyer

image

zssherman:
Looks like just some linting errors, besides that, looks good to me

zssherman (approved):

AdamTheisen:
Hold off on merging just yet.  I noticed that the 1d timeseries is not the right color.  Need one more fix

zssherman:
Sounds good!

AdamTheisen:
Actually, I think this is good to go.  There's some more complication if we change the colormap for the line plots and it might not play well with many lines.  @zssherman if you're good with this, we can merge in.  Thanks!

zssherman:
Sounds good! merging

PR #615, opened 2023-02-02
CI: Switching to codecov

zssherman:
Coveralls is going to a pay model. Switch from coveralls to codecov.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -17,6 +17,7 @@ coverage:
     project:
       default:
         target: 95
+        threshold: 2%
         informational: true
     patch: off
     changes: off
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,22 @@
+codecov:
+  require_ci_to_pass: no
+  max_report_age: off
+
+comment: false
+
+ignore:
+  - 'act/tests/*'
+  - 'act/*version*py'
+  - 'setup.py'
+  - 'versioneer.py'
+
+coverage:
+  precision: 2
+  round: down
+  status:
+    project:
+      default:
+        target: 95
+        informational: true
+    patch: off
+    changes: off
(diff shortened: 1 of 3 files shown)


AdamTheisen:
@zssherman so we lost the test to ensure that the code coverage did not decrease by a certain threshold.  Do we have that ability with Codecov?

zssherman:
Good point, I'll dig through the docs to see if I can find that

codecov.yml
```diff
@@ -0,0 +1,22 @@
...
+  - 'setup.py'
+  - 'versioneer.py'
+
+coverage:
+  precision: 2
+  round: down
+  status:
+    project:
+      default:
+        target: 95
```
mgrover1, line 19:
```suggestion
        target: 95
        threshold: 2%
```

mgrover1 (changes requested):
One small suggestion - can we add in the "threshold" setting which allows up to a 2% decrease in code coverage in a PR?

mgrover1:
@AdamTheisen @zssherman I suggested the configuration addition that sets the % decrease threshold

codecov.yml
zssherman: Ah perfect so there is a threshold! Changing now

mgrover1: You betcha!

zssherman: Has been changed

AdamTheisen:
Thanks @mgrover1 !  How did I miss that.  I assume that the builds will fail if the coverage decreases right or is there a way to add a standalone check like we had with coveralls?

mgrover1:
> Thanks @mgrover1 ! How did I miss that. I assume that the builds will fail if the coverage decreases right or is there a way to add a standalone check like we had with coveralls?

We would not be able to do this in a separate step/check.

mgrover1 (approved):

Issue #402, opened 2022-03-01
Groupby Plots


AdamTheisen:
It would be beneficial to add a function that will plot data grouped by certain periods of time.  For instance, the following is fake data that's grouped and plotted by year.  Goal would be to provide generic function to handle groupby's for day, month, year.  

Year Grouping
![image](https://user-images.githubusercontent.com/13421074/156261678-622b47c7-987d-4b6e-8ddb-e27b297ba638.png)

Month Grouping
![image](https://user-images.githubusercontent.com/13421074/156262963-15ff4250-a576-42d3-bf11-9d1a3fd134b7.png)

zssherman:
@rcjackson Has there been any progress on this? See you self-assigned

rcjackson:
@zssherman It fell of my priority list given the craziness of last summer, but I just put it back on and will start to look at it at the end of this week

From: Zach Sherman ***@***.***>
Date: Wednesday, February 15, 2023 at 9:33 AM
To: ARM-DOE/ACT ***@***.***>
Cc: Jackson, Robert ***@***.***>, Mention ***@***.***>
Subject: Re: [ARM-DOE/ACT] Groupby Plots (Issue #402)

@rcjackson<https://github.com/rcjackson> Has there been any progress on this? See you self-assigned

zssherman:
@rcjackson Cool! Let me know if there is anything I can do to help with this

AdamTheisen:
Closing out since @rcjackson did some great work on this in PR #621

PR #621, opened 2023-02-17
ADD: Groupby plotting capability.

rcjackson:
Addresses #402. I have now added a wrapper around the Display object that allows the user to group their display by time periods (right now it's (year, month, day, hour, minute, second)). 

The groupby can be initiated with any display object using groupby = display.group_by('year') if you want to group the plot by year. Then you would use groupby.plot_group('plot_func_name', datastream, keyword arguments for the plotting function in the Display object)

For example, this does wind rose data plots grouped by day for the data rose plot example:
obj = act.io.armfiles.read_netcdf(act.tests.EXAMPLE_MET_WILDCARD)

# Create Plot Display
display = act.plotting.WindRoseDisplay(obj, figsize=(15, 15), subplot_shape=(3, 3))
groupby = display.group_by('day')
groupby.plot_group('plot_data', None, dir_field='wdir_vec_mean', spd_field='wspd_vec_mean',
                   data_field='temp_mean', num_dirs=12, plot_type='line')
image


mgrover1 (approved):
Great addition! I like the class used here, and the tests/example you put together.

zssherman:
@rcjackson The test failure is unrelated and fixed in Ken's PR. But yeah this looks great!

zssherman:
@rcjackson Is there anyway to fix the example plot, the overlap of title and windrose top marker?

mgrover1:
a `plt.tight_layout()` should fix it?

zssherman:
Rerunning tests as kens PR should fix this now.

examples/plotting/plot_days.py
```diff
@@ -0,0 +1,26 @@
+"""
+Calculate and plot wind rose plots separated by day.
+-----------------------------------------------------
+
+Example of how to read in MET data and plot histograms
+of wind speed and temperature grouped by day.
+
+Author: Adam Theisen
```
AdamTheisen, line 8: @rcjackson I think you need to update the author on this one.

zssherman:
Resetting tests to match changes in Ken's PR

AdamTheisen:
Looks like a cropscape test is failing so they must have updated the DB. We can change that test to be = Soybeans.

rcjackson:
Fixed the plotting example to move the axes ticks inside the rose.

AdamTheisen:
@rcjackson I'm running into some issues when I try and test this out on plotting daily temperatures.  Code is 

```
obj = act.io.armfiles.read_netcdf(act.tests.sample_files.EXAMPLE_MET_WILDCARD)
print(obj['time'].values)
display = act.plotting.TimeSeriesDisplay(obj, subplot_shape=(3, 3))
groupby = display.group_by('day')
groupby.plot_group('plot', None, field='temp_mean')
plt.show()
```

But the only plot that has data is the first one.  Am I missing something?
image

AdamTheisen:
Note, I got the same when trying to do a month groupby on all of 2022 data.

image

rcjackson:
I think the default behavior of TimeSeriesDisplay to set all x axes limits the same was causing it. In this case, we don't want them to be the same or they would all just be for the first time period. I put some code to override that behavior in the GroupBy when it detects a TimeSeriesDisplay to set to the min and max time for each group.
image

zssherman (commented):
@rcjackson Do we know if this will occur elsewhere in other plotting modules?

act/plotting/plot.py
```diff
@@ -278,3 +279,126 @@ def add_colorbar(self, mappable, title=None, subplot_index=(0,), pad=None,
...
+        """
+        self.display = display
+        self._groupby = {}
+        num_groups = 0
+        datastreams = list(display._obj.keys())
+        for key in datastreams:
+            self._groupby[key] = display._obj[key].groupby('time.%s' % units)
+            num_groups = max([num_groups, len(self._groupby[key])])
+
+    def plot_group(self, func_name, dsname, **kwargs):
```
mgrover1, line 320: Is there anyway we can automatically fill the dsname to be None?

rcjackson:
> @rcjackson Do we know if this will occur elsewhere in other plotting modules?
None of the other modules use time as an x-axis, so it should not affect them.

act/plotting/plot.py
rcjackson: Done in new commit!

zssherman (approved):

zssherman:
Looks good, if no one else has anything to add, will merge this.

AdamTheisen:
Just tested and looks good!  Thanks for the updates @rcjackson !  @zssherman I think we're good to go

PR #624, opened 2023-02-21
FIX: Groupby plots will now group timeseries subplots by 

rcjackson:
The groupby plot functionality now groups subplots by year for timeseries plots if there are multiple years present in a subplot. This makes it able to put different years of data on different curves like below.
image


zssherman (approved):

mgrover1 (approved):

mgrover1:
Rerunning the 3.7 test

PR #637, opened 2023-03-11
Xlim single value warning

### LINKED ISSUE #629: Attempting to set identical low and high xlims
* ACT version: 1.4.0
* Python version: 3.8.13
* Operating System: Linux

### Description
When processing in an automated queue we sometimes process when there are few time steps. The process does not fail but does produce a warning. 

/.conda/envs/dqo-base/lib/python3.8/site-packages/act/plotting/timeseriesdisplay.py:214: UserWarning: Attempting to set identical low and high xlims makes transformation singular; automatically expanding.

I want to add a check when setting xlim values to ensure start and end are not the same. If so automatically expand the xlim to exclude the warning.

kenkehoe:
This will check if the xlim is attempting to set a single value for start and end limit of plot. If so will expand by 1 second on either side so there is no warning about setting the limit to a single value.

- [ X] Closes #629 
- [X ] Tests added


zssherman (changes requested):
The changes looks good to me, the only thing I suggest is backwards compatibility with the colormap change. I would do:

```
    try:
        matplotlib.colormaps.register(name=full_name, cmap=cmap, force=True)
    except AttributeError:
        matplotlib.cm.register_cmap(name=full_name, cmap=cmap)
 ```

kenkehoe:
Windows-3.7 is taking a very long time. Is this to be expected?

kenkehoe:
OK should we restart the testing?

zssherman:
Yep restarting now

PR #638, opened 2023-03-16
Correctly handling PPI derived winds when the function returns all NaNs

kenkehoe:
There is an issue when processing DL PPI to derive winds when the winds are not calculated. The returned values are np.nan values instead of xarray Dataset. Merging np.nan using xarray causes an error. This update checks data type and only attempts to merge Datasets.


mgrover1 (approved):
Good catch @kenkehoe !

PR #640, opened 2023-03-17
Dod example

### LINKED ISSUE #591: Add option to include required global attributes when writing to netCDF
Some people are planning to use ACT to generate official datastreams and need to have required global attributes set. We should help them add the correct attributes with correct values.

AdamTheisen:
- [x] Closes #591 

I think this would suffice for taking care of #591 .  I think it is easy enough to update/add attributes using xarray that ACT does not need the added functionality to set these specific attributes for the users.


zssherman:
Sounds good! I'll take a look in a short bit, picking up my car from finally being repaired lol

zssherman (approved):

PR #641, opened 2023-03-17
Updated to new required keyword in stats.mode(). 

kenkehoe:
There is a warning from stats.mode() indicating that the need to indicate if a single value dimension should be dropped from the returned values. This PR explicitly sets the keyword to remove the warning and return a scalar when we desire a scalar.

Needed to add a try/except for the numpy 1.10 to 1.11 transition period. The keyword only exists in Numpy 1.11 and newer so for backward compatibility need to catch the error and use previous syntax.

- [X ] PEP8 Standards or use of linter


zssherman (approved):
Looks reasonable to me!

Issue #606, opened 2023-01-20
improving the function of the current skew-T plot


Aifang-Zhou:
I am working on the VAP SondePram. We'd like to display the values of CAPE/CIN/LCL/LFC/LNB on the skew-T plot, and these values are supposed to reflect the calculation results of the VAP. Could you improve the skew-T display so that we can realize this function? Please let me know if you need any information.

AdamTheisen:
Thanks for the issue @Aifang-Zhou!  We might want to think about this in 2 ways

First Option: Provide a keyword to add these variables using MetPy/Sonde Stability function in ACT
Second Option: Allow the user to pass in a dictionary of keys and values to display in a set area that would be populated with the first option.

Additionally, I wonder if we could further improve the Skew-T plots with the option to add a hodograph?

@mgrover1 @zssherman any thoughts here?

zssherman:
@AdamTheisen I think either way should be fine, whichever is easier for the user to input those fields.

AdamTheisen:
Putting this here for the record.  We could follow this layout and then under the hodograph, we can add results from the stability retrieval or have the option for the user to pass in a dictionary of info to display.

https://unidata.github.io/python-training/workshop/Skew_T/skewt-and-hodograph/

AdamTheisen:
I would like this to stay open to get feedback from @Aifang-Zhou and plan for future improvements with the Skew-T display

zssherman:
I'm like, I don't remember closing this one lol, but remembered that the closes issue in a pr merge closes it

Aifang-Zhou:
Thank you, @AdamTheisen  and @zssherman, for the feedback and the example! 

May I ask you a question about the title of skewt plots? 

Suppose I use the data from "sondewnpn.b1" to create a skewt plot, and use data from "sonparam" to display the texts, then the title of the skewt plot is like "sondewnpn.b1 on mm/dd/yy". Since the plot is supposed to be created by VAP SONDEPARAM, we want to change the title to "sondeparam.c1 on mm/dd/yy". Is there a function I can call to do this?

zssherman:
@Aifang-Zhou No worries, and you can override and set your own title with the set_title parameter in the plotting function. If your looking to mimic the title creation but with the vap datastream, can always plug in this to set_title:
set_title = ' '.join(["sondeparam.c1", field, 'on', dt_utils.numpy_to_arm_date(ds.time.values[0]),])
I believe should work.

[...]

PR #644, opened 2023-03-19
Skewt

### LINKED ISSUE #607: Make SkewT Display More Flexible with Matplotlib
While taking a look at #606 , I noticed that there is not currently a way to feed in a user-defined figure or axes to the SkewT Display. This would be an important feature to add, enabling users to customize their plots more, and interact with other displays.

AdamTheisen:
- [x] Closes #607 and #606 
- [x] Tests added


mgrover1:
@AdamTheisen it does appear these are remaining SkewT plots... the temperature/pressure lines should be skewed with decreasing pressure.

Can you check to make sure the axes are set to be skewed?

Here is what the test plot looks like:

Screenshot 2023-03-20 at 8 16 05 AM

And what it should it look like (not the same sounding, but notice the temperature/pressure lines)
image

mgrover1:
I think the key here is making sure the axis type is still `SkewXAxes` and `SkewYAxes`

https://github.com/Unidata/MetPy/blob/main/src/metpy/plots/skewt.py#L162

AdamTheisen:
@mgrover1 @zssherman I have a fix but it would require matplotlib >= 3.7 for this plot.  Is this acceptable?

AdamTheisen:
I can't seem to see the updated baseline plots on here but these are it
test_enhanced_skewt_plot

test_enhanced_skewt_plot_2

zssherman:
@AdamTheisen Pinning matplotlib I think would be fine if it works with all the CI python versions

AdamTheisen:
@zssherman okay, I'll update the requirements files when I get a chance.  The error right now looks like what I was getting when I didn't have matplotlib 3.7

mgrover1:
Yeah, metpy does not support Python 3.7, so there will not be any fixes for that error

AdamTheisen:
@mgrover1 @zssherman  looks like the latest error is saying matplotlib is not valid for 3.7.  Would we want to drop 3.7 testing?

mgrover1:
Yeah, and I think we want >=3.8 since 3.7 isn't supported

mgrover1 (approved):

mgrover1:
Tests are passing! Thanks for putting this together @AdamTheisen

AdamTheisen:
Nothing else is planned at this time!  Will merge in

PR #645, opened 2023-03-20
CI: Update, remove Python 3.7 and add 3.11.

zssherman:


mgrover1 (approved):

zssherman:
Might honestly fix our stall issues as well as they often occur for 3.7

PR #668, opened 2023-05-10
ADD: Add Sodar reader.

zssherman:
Adding a sodar reader for the atmos site. Reader reads in .mnd files. Will add tests and an example to this PR as well.

- [x] Tests added


act/io/sodar.py
```diff
@@ -0,0 +1,196 @@
...
+import fsspec
+import numpy as np
+import pandas as pd
+
+from act.io.noaapsl import filter_list
+
+
+def read_sodar(filepath):
+    """
+    Returns `xarray.Dataset` with stored data and metadata from a user-defined
```
mgrover1, line 18: Can you specify what type of sodar is used here? With a link to the instrument docs?

mgrover1 (changes requested):
One small change - otherwise, this looks great!! Thanks for your hard work here @zssherman

act/io/sodar.py
zssherman: Done

mgrover1 (approved):
Great! Thank you @zssherman

PR #671, opened 2023-05-11
MNT: Update .gitignore to be more robust with downloaded test files.

zssherman:


mgrover1 (changes requested):
Can you add a small bit in the contributors guide that data files from the `download_data` function will need to be added to the git ignore?

Issue #548, opened 2022-11-01
Have both QC and DQR  classes apply filtering work done to the attributes section


caghili:
When you load a file into xarray, you see  the attributes of the file. I was wondering if after applying
 QC  obj.clean.cleanup() or 
DQR  add_dqr_to_qc(obj, variable=None, assessment='incorrect,suspect', exclude=None, include=None, normalize_assessment=True,  cleanup_qc=True)   

these could be up dated to reflect that. So once you save out the file then reload it you can see the filtering done on a particular datastream.

![qc_dqr](https://user-images.githubusercontent.com/117209377/199305085-c1080c2c-8acf-43fe-a017-705c02a97b44.png)

mgrover1:
Thank you for this suggestion - is this something you would be interested in contributing?

caghili:
> Thank you for this suggestion - is this something you would be interested in contributing?

I would, was sick the other day, however I'm very busy supporting other groups. If I were to support it'd be the odd hour here and there (I could see how things progress and get to know this group).

zssherman:
@AdamTheisen This was handled in the most recent qc PR I believe? I'll close, but can reopen if need be,

PR #675, opened 2023-05-23
QC Updates

### LINKED ISSUE #547: Ability to see DQR Report link with  add_dqr_to_qc (or an additional class to the QC module)
Something along these lines, having the ability to get the DQR report for a datasteam of interest via python  would cut down on some time.  like somehow spit out a link after add_dqr_to_qc. Since sometimes datastreams have no dqr reports.

AdamTheisen:
Addressing #547 and #548 but came across an issue in the write_netcdf function that causes issues with ARM's base_time string attribute.

- [x] Closes #547 and #548 
- [x] Tests added


AdamTheisen:
I really am not sure where all the commits come from in here.  I just opened this branch and started working on this today.

act/qc/qcfilter.py
```diff
@@ -1050,6 +1050,43 @@ def datafilter(
...
+                rm_assessments = [rm_assessments]
+            if rm_assessments is not None:
+                for assessment in rm_assessments:
+                    if assessment in flag_assessments:
+                        index = [i for i, e in enumerate(flag_assessments) if e == assessment]
+                        for ind in index:
+                            comment = ''.join(['act.qc.datafilter: ', flag_meanings[ind]])
+                            if 'history' in self._ds[var_name].attrs.keys():
+                                self._ds[var_name].attrs['history'] += '\n' + comment
+                            else:
```
zssherman, line 1087: More for clarification if rm_tests exist and appends to the comment, will rm_assessments then append to that as well? I think I understand what's happening here

zssherman:
A couple comments, one change possibly.  Tests seem to fail as rm_tests in an int. Should it be i in range(rm tests) or is rm_tests going to be a list datafilter changes?

AdamTheisen:
Looks likes some updates with MetPy calcs so will handle in the morning

act/qc/qcfilter.py
AdamTheisen: Yep, essentially comment is just a dummy variable that is being appended to the history attribute for both rm_assessments and rm_tests.  It's a little more complicated for rm_assessments though since it could just be "Bad" and we need to find all the bad tests.

AdamTheisen:
@zssherman I fixed the bug for when rm_tests is a single int.  I change it to a list and then iterate through.

AdamTheisen:
@zssherman if I addressed your comments go ahead and merge this on in!  If not, let me know.

PR #677, opened 2023-05-23
refactor: changing to allow user to pass layout of mosaic subplots ex…

cgodine:
…plicitly as a keyword argument. 

Change will give user more flexibility on how the enhanced skew-t subplots are laid out on the image. 

### DIFF
--- a/act/plotting/skewtdisplay.py
+++ b/act/plotting/skewtdisplay.py
@@ -694,6 +694,7 @@ class SkewTDisplay(Display):
         uv_flag=False,
         dsname=None,
         figsize=(14, 10),
+        layout=None,
     ):
         """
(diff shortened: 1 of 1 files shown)


zssherman (approved):
@cgodine Thanks for the PR! looks good to me, will see what the tests do.

zssherman:
@cgodine More I thought about it, can we set the default to stay constrained, with the docstring stating that. But still keep the ability for the user to define what they want?

cgodine:
> @cgodine More I thought about it, can we set the default to stay constrained, with the docstring stating that. But still keep the ability for the user to define what they want?

Yeah, I think with the way testing is set up that might be an easier fix.

zssherman (approved):

zssherman:
@cgodine Thanks for making the changes, merging!

PR #678, opened 2023-05-26
ENH: Potential fix for applying DQRs to non time dimensioned variable…

### LINKED ISSUE #530: Handling Incorrect ARM DQRs when applied with ACT function
We have implemented a function to query the ARM Data Quality Report database to return time periods when data is flagged. There are three levels of flagging within the DQRs, with two of them able to replace the variable values with NaN. ARM has a lot of DQRs where the list of variables is not accurate, and is over listing the affected variables.

For example a DQR may indicate the instrument is malfunctioning and select all variables in the netCDF file. While the data is incorrect the location, time, and QC variables are correct. Currently we loop over variables and apply the DQR to variables listed in the DQR. Some variables listed are scalars which is causing an exception that we catch and handle. Handling the exception has a print statement indicating the scalar variable could not be indexed.

We should discuss a better way to handle these variables as the print statement may confuse users when there is nothing wrong with the scalar variables. A couple of options include:

1) Correctly handling the replacement of data values with NaN for the scalar variables
2) Setting a keyword to enact the print statement, but set to False as default
3) Skip over these most likely incorrectly listed variables.

An example output for sgpaosccn2colaE13.b1 on 20170903 using D170905.3
```
Skipping 'eta_lookup_table' DQR application because of IndexError
Skipping 'base_time' DQR application because of IndexError
Skipping 'alt' DQR application because of IndexError
Skipping 'droplet_size_bounds' DQR application because of IndexError
Skipping 'lon' DQR application because of IndexError
Skipping 'lat' DQR application because of IndexError
```

AdamTheisen:
…s or scalars

- [x] Closes #530 

### DIFF
--- a/act/qc/arm.py
+++ b/act/qc/arm.py
@@ -146,8 +146,12 @@ def add_dqr_to_qc(
             starttime = np.datetime64(dt.datetime.utcfromtimestamp(int(line[1])))
             endtime = np.datetime64(dt.datetime.utcfromtimestamp(int(line[2])))
             ind = np.where((time >= starttime) & (time <= endtime))
+
             if ind[0].size == 0:
                 continue
+
+            if 'time' not in ds[var_name].dims:
+                ind = np.where((ds[var_name].values == ds[var_name].values) | (np.isnan(ds[var_name].values)))
 
             if dqr_no in dqr_results.keys():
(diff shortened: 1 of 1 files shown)


zssherman:
@AdamTheisen Sounds good!

zssherman (approved):
Looks good to me!

zssherman:
Rerunning stalled jobs

PR #681, opened 2023-06-07
ENH: Bug fix for setting tests when it's a scalar

AdamTheisen:
I found an error with the previous update that I was trying to get to.  This should alleviate issues with trying to set qc values for scalar data.  


zssherman (approved):

mgrover1:
@AdamTheisen can you add a test here?

AdamTheisen:
@mgrover1 @zssherman one test added in which helped find a bug so we should be good to go when tests pass.

mgrover1 (approved):
Great! Thank you @AdamTheisen !

Issue #459, opened 2022-05-05
Need a new plot module to allow for non-time plotting on X-axis


AdamTheisen:
There's been a request to allow for non-time plotting on the xaxis, like solar zenith or other variables.  A new plotting module for x-y plots and not specifically time series would be beneficial.

AdamTheisen:
@cgodine did you ever get something going for this one?  If not, I can always take a first pass at it.

zssherman:
I think this was tackled in #683 can probably close.

AdamTheisen:
@zssherman I missed your comment but agree!  This was covered in scatter plot feature.

PR #683, opened 2023-06-12
Comparison Display 

jrobrien91:
New Display object to compare two datastreams through the use of scatter and violin plots.
- Scatter plots allow for plotting of non-time variable along the x-axis
    - complement to 2D Histogram plot within HistogramDisplay
- Violin plots allow a view of the data distribution and Interquartile range

- [X] Closes/First Cut at #459 
- [X] Tests added


zssherman:
@jrobrien91 I'll take a look tomorrow

AdamTheisen:
In meetings all day but can look and test out tomorrow

AdamTheisen:
@jrobrien91 this is looking good!  I want to get feedback from @mgrover1 and @zssherman but I'm wondering if ComparisonDisplay is the best class name for these.  I.e. would ScatterDisplay be better for the scatter plot and something like DistributionDisplay for the violin or something like that?  

Which leads me to wondering if we need to rename the HistogramDisplay to DistributionDisplay?  Thoughts?

jrobrien91:
@AdamTheisen but I agree, `ComparisonDisplay` was always an awful name and another name more helpful would be appreciated

mgrover1:
I like the idea of consolidating this into `DistributionDisplay`

zssherman:
@jrobrien91 @AdamTheisen @mgrover1 I second a distribution display with  the underlying functions between this all within the module.

zssherman:
If were all in agreement, I say make the changes and I can review it once its done.

AdamTheisen:
@zssherman @mgrover1 do we need to keep the HistogramDisplay option around with a deprecation warning for a bit or do you think we can go forward with the change and do a minor release?

mgrover1:
Let's put in a deprecation warning and remove in the next major release!

AdamTheisen:
Sounds good @mgrover1 !

One last comment on the scatter plot.  It might be very helpful to be able to pass in a dsname for the x and y variables separately.  For example, I might want to compare the temperature from the MET and MAWS systems.  Is that doable?  Happy to help with these other updates as well.

zssherman (approved):
Everything looks good to me @jrobrien91 @AdamTheisen. I have no comments, I think the depreciation warning is good as well!

mgrover1 (approved):
Looks good to merge! Thank you for the changes @jrobrien91 @AdamTheisen

PR #689, opened 2023-06-19
Adding SURFRAD Capabilities

AdamTheisen:
- [x] Tests added


AdamTheisen:
Looks like there are some errors with the PySP2 tests

mgrover1:
@AdamTheisen it appears that there is some 0 shaped array being multiplied within the try/except block...
```python
if np.isfinite(base):
            try:
                height = data_fit.max() - base
                pos = np.argmax(data)
            except ValueError:
                height = np.nan
                pos = np.nan
    
            try:
                start = np.where((data - base) < 50)[0]
>               if start == []:
E               ValueError: operands could not be broadcast together with shapes (14,) (0,)
````

zssherman:
Nothing changed within the pysp2 repo itself, hmm so curious why it's failing now

mgrover1:
> Nothing changed within the pysp2 repo itself, hmm so curious why it's failing now

pysp2 does not currently have a nightly build running/testing with dependencies, so I suspect is has to do with some upstream library cutting a new release

mgrover1:
@zssherman I suspect numpy is the culprit here... https://github.com/numpy/numpy/releases/tag/v1.25.0

zssherman:
@mgrover1 @AdamTheisen Looks like we need a change upstream. So when it checks the start for empty list, check if its a numpy array

mgrover1:
> @mgrover1 @AdamTheisen Looks like we need a change upstream. So when it checks the start for empty list, check if its a numpy array
> 
> ```
> try:
>     start = np.where((data - base) < 50)[0]
>     if start == []:
>         start = np.nan
> ```

Upstream as in pysp2?

zssherman:
@mgrover1 I believe so

mgrover1:
Ahh the good old `method=nearest` argument - good catch @AdamTheisen !

AdamTheisen:
> Ahh the good old `method=nearest` argument - good catch @AdamTheisen !

Found another issue with the plot and the text showing up as expected while looking so it was a good error to help improve the example!

mgrover1 (commented):
@AdamTheisen - does their documentation include anything about the `standard_name` for the variables? I think that would be helpful include here if that is the case.

AdamTheisen:
> @AdamTheisen - does their documentation include anything about the `standard_name` for the variables? I think that would be helpful include here if that is the case.

The only documentation they really have is what's in the [README](https://gml.noaa.gov/aftp/data/radiation/surfrad/Boulder_CO/README) file.  They don't have anything about standard_names.

mgrover1:
> > @AdamTheisen - does their documentation include anything about the `standard_name` for the variables? I think that would be helpful include here if that is the case.
> 
> The only documentation they really have is what's in the [README](https://gml.noaa.gov/aftp/data/radiation/surfrad/Boulder_CO/README) file. They don't have anything about standard_names from what I can see.

Gotcha - any chance you could take a look at this table
https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html

And see which ones would map? One benefit of using a `cf-standard` name for as many variables as possible is the ability to use tools such as [`cf-xarray`](https://cf-xarray.readthedocs.io/) to generically compare different datastreams using these community standards...

zssherman (approved):

zssherman:
Looks good to me, tests are now passing!

AdamTheisen:
Just as a note, I did not get a chance to add in standard names yet. Was hoping to look into that tomorrow

mgrover1:
Sounds good - I think that makes sense for separate PR 😄

PR #696, opened 2023-07-07
Random updates

AdamTheisen:

### DIFF
--- a/act/corrections/mpl.py
+++ b/act/corrections/mpl.py
@@ -131,7 +131,6 @@ def correct_mpl(
         co_ap = np.transpose(co_ap)
 
     x_ap = ds[cross_pol_afterpulse_var_name].values
-
     # Fix dimentionality if backwards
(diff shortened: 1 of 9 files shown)


mgrover1 (commented):
@AdamTheisen - this will change the API - should we create a v2 tag and pull request to that instead of main?

zssherman:
@mgrover1 @AdamTheisen its a duplicate file and one has a warning, the api won't change on the user's end besides a new warning, and the replacement will also exist.

AdamTheisen:
@mgrover1 what @zssherman said!  :). Shouldn't change the API at the moment.  Figured that we would do some cleanup with V2.

mgrover1:
Okay! Great! I wasn't sure if we wanted to create that separate branch for now - but it sounds like that is a no, so I can go ahead and approve!

mgrover1 (approved):

PR #700, opened 2023-07-27
ADD: 'extract_arm_file_info' to harvest info from lists of ARM

isilber:
filenames without loading the files, thereby saving header reading and processing times

### DIFF
--- a/act/io/__init__.py
+++ b/act/io/__init__.py
@@ -17,6 +17,7 @@ __getattr__, __dir__, __all__ = lazy.attach(
             'read_netcdf',
             'check_if_tar_gz_file',
             'read_mmcr',
+            'extract_arm_file_info',
         ],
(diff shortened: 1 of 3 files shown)


zssherman:
@isilber Thanks for the PR! Would it be possible to add a unit test for the new function?

AdamTheisen:
Just as a note, I do think it would be beneficial to restructure this like how the DQ Office uses regular expressions to extract the corresponding data.  That would enable it to be more robust and work across all ARM cases.  @kenkehoe thoughts?

kenkehoe:
Sounds good to me. Shall I put our filename parser into ACT?

AdamTheisen:
@isilber thanks again for this PR!  We ended up merging in the code in #708 which performed this functionality.  Let us know if you see any issues and we'll hope tp get you in as a contributor/co-author soon!

PR #709, opened 2023-09-14
FIX: Upgrade various actions and ensure proper env name is set

mgrover1:
Upgrades various actions and makes sure the proper variables are set for the micromamba action


AdamTheisen (approved):

AdamTheisen:
Thanks @mgrover1 !  Merging in

PR #712, opened 2023-09-23
Fix `qcfilter.add_test` index

maxwelllevin:
This PR fixes an issue where calling `ds.qcfilter.add_test(...)` with `index=np.array([False])` or `index=np.array([True])` would set the bit to the opposite of what it should be.

- [X] Tests added


AdamTheisen (approved):
Thanks @maxwelllevin, this looks good.  I'll wait for @mgrover1 and @zssherman to add a second review and then we can merge in.

zssherman (approved):
Looks good to me as well! Merging

PR #715, opened 2023-09-26
FIX: Warning fixes.

zssherman:
* Sets datetime precision to ns for xarray to handle due to pandas.
* Sets time encoding to time bounds encoding if time has no encoding
* Fixes multiindex from pandas with xarray handle
* Add r for regex calls


zssherman:
Will fail, I believe pandas pinning in CI is causing a lower Xarray version which does not have attribute i'm using here.

PR #717, opened 2023-09-27
CI: Pandas remove

zssherman:
Had to reopen PR as PR wasn't updating to new commits (First for everything). Also add timeout for get_ord stalling

### DIFF
--- a/act/tests/test_plotting.py
+++ b/act/tests/test_plotting.py
@@ -333,7 +333,7 @@ def test_skewt_plot_spd_dir():
         matplotlib.pyplot.close(skewt.fig)
 
 
-@pytest.mark.mpl_image_compare(tolerance=80)
+@pytest.mark.mpl_image_compare(tolerance=81)
 def test_multi_skewt_plot():
 
     files = glob.glob(sample_files.EXAMPLE_TWP_SONDE_20060121)
--- a/continuous_integration/environment_actions.yml
+++ b/continuous_integration/environment_actions.yml
@@ -22,7 +22,7 @@ dependencies:
   - pytest-cov
   - pytest-mpl
   - coveralls
-  - pandas<2.0
+  - pandas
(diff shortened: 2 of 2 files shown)


zssherman:
@AdamTheisen @mgrover1 Thoughts on adding a pytest-timeout module to pass on the get_ord tests if it goes longer then 5 minutes? This happens every PR for atleast one of the suites

mgrover1:
> @AdamTheisen @mgrover1 Thoughts on adding a pytest-timeout module to pass on the get_ord tests if it goes longer then 5 minutes? This happens every PR for atleast one of the suites. Granted I don't know why the io_write stalls for some tests sometimes

I think this brings up a bigger issue here - we should add sample files to test on and not rely on hitting these servers since a suite of python versions + machines will often overload their bandwidth. I would rather have a single file we test against that we know is on a system with improved bandwidth than pass when things timeout

AdamTheisen:
@mgrover1 @zssherman I thought we added a timeout to that test already but maybe not. I think we should discuss the file part more.  We've had cases in the past where the API updated and it was only through these tests that we would find it. If we are going off a static file we could be using outdated formats. Thoughts?

zssherman:
@AdamTheisen @mgrover1 Yeah I agree on testing the API so we know if something changes. However I also agree with bandwidth issues as well, is there a way to ping the API and not download the file to just see if the path is valid? Also we did not have the timeout, also I believed it was suppose to skip it if it took to long from the documentation, but it looks like it failed the testing suite which isn't what I wanted... so the timeout might not work anyways

zssherman:
@AdamTheisen I removed the pytest timeout stuff, are we good to merge this? This will fix the CI for some of the other PRs I have

mgrover1 (approved):

AdamTheisen (approved):
@zssherman please go ahead!

PR #720, opened 2023-10-05
FIX: Update versioneer.py for python 3.12 support.

zssherman:


mgrover1 (approved):

mgrover1:
Thanks for the quick fix here @zssherman

zssherman:
Geoplot test is giving different results for the array each time it runs for ubuntu....

mgrover1:
@zssherman - the default figsize changes every time. The best way forward is to manually set the figsize, I added a fix here with an update to the image to validate.

mgrover1:
@zssherman - tests are good to go! The base image just needed to update I believe

PR #722, opened 2023-10-09
ENH: Bug fix for geoplot example

AdamTheisen:
There's an issue with the cartopy Stamen that is causing errors.  While we sort through that, setting the stamen to terrain and the tile to 9 seems to produce consistent results.

### DIFF
--- a/act/io/armfiles.py
+++ b/act/io/armfiles.py
@@ -17,6 +17,7 @@ import tarfile
 import tempfile
 import warnings
 
+from cftime import num2date
 import numpy as np
 import xarray as xr
 import datetime as dt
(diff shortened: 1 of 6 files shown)


zssherman:
Thanks @AdamTheisen for the fix!

zssherman:
Looks like the error is a syntax error

mgrover1 (approved):
Great!!

AdamTheisen:
It looks like one test failed so maybe this is not a permanent solution.  Rerunning to see what it does.

AdamTheisen:
@zssherman @mgrover1 there was one last error with the geoplot test.  I'm wondering if I should update the geoplot test to follow the example.  Thoughts?

PR #723, opened 2023-10-09
Improvements for the secondary y axis functionality

### LINKED ISSUE #163: Secondary Y axis
@rcjackson @kenkehoe I was looking at adding in a feature for working with a secondary y axis on the timeseries plots, but it is a little more in depth than I thought, mainly due to the set_yrng function.  We rely on the subplot indices to set a lot of things and that currently does not include anything about a secondary y-axis.  I could put in a work around for now that I think will work, but it's not ideal long-term.  Could we adjust the subplot indices at all or are there any other options you could think of?

AdamTheisen:
This changes the functionality of what's returned as part of the display.axes object.  For the TimeSeriesDisplay and DistributionDisplay it will now return a 2D array for each plot (left and right axis object) instead of a 1D array of just the left axis object for each plot.  This will break conventions so might be something we consider for V2.0.0

- [x] Closes #163 
- [x] Tests added


zssherman:
@AdamTheisen Looks fairly reasonable to me. Tests are failing due to that geoplot issue.

mgrover1 (approved):

zssherman (approved):

PR #729, opened 2023-10-10
FIX: Remove verify warning, by setting verify to true.

zssherman:


AdamTheisen:
@zssherman let me test this out a bit.  I know in the past, this caused issues with their API.

AdamTheisen:
@zssherman works fine now but we might want to update the function description as we did note that we were setting that to False

`Note, the requests library
    is indicating a bad handshake with the server so 'verify' is currently
    set to False which is unsecure`

Wondering if we need an issue to add a citation as well... Thoughts?
`USDA National Agricultural Statistics Service Cropland Data Layer. {YEAR}. Published crop-specific data layer [Online]. Available at https://nassgeodata.gmu.edu/CropScape/ (accessed {DATE}; verified {DATE}). USDA-NASS, Washington, DC.`

zssherman:
Oh yep! I missed that note, I can remove it, but agreed I think a citation would be good

zssherman:
@AdamTheisen I updated the PR to remove the note and added the reference, let me know if it looks good!

AdamTheisen (approved):

AdamTheisen:
@zssherman I think this is good to go!  Thanks!

PR #734, opened 2023-10-13
ADD: Adding new example for working with ARM QC and Xarray Transforma…

AdamTheisen:
…tions.  Updating one example to follow ds convention, and fixing warn message for armfiles

### DIFF
--- a/examples/workflows/plot_qc_transforms.py
+++ b/examples/workflows/plot_qc_transforms.py
@@ -2,10 +2,9 @@
 Transformations and QC
 ----------------------
 
-Built-in transformations using xarray do not
-handle ARM quality control appropriately.  This
-example shows how a user can apply QC prior to
-performing transformations.
+Built-in transformations using xarray are not
+quality-control aware.  This example shows how
+a user should apply QC prior to performing transformations.
 
 """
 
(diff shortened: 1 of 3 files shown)


examples/workflows/plot_qc_transforms.py
```diff
@@ -0,0 +1,49 @@
+"""
+Transformations and QC
+----------------------
+
+Built-in transformations using xarray do not
+handle ARM quality control appropriately.  This
```
mgrover1, line 6: I re-phrase as they are not quality-control aware. It is not necessarily something that xarray should inherently do

mgrover1 (changes requested):
One minor comment - otherwise, this looks great!

examples/workflows/plot_qc_transforms.py
AdamTheisen: Good point!  Will update!

AdamTheisen:
@kenkehoe just wanted to ping you on this example.  Was something like this what you were thinking?

zssherman (approved):

AdamTheisen:
@mgrover1  does this look good now?

mgrover1 (approved):
Thanks for making the changes here!

kenkehoe:
@AdamTheisen  overall I think it looks good. I didn't actually run the code so I don't know what the plot looks like. Does the updated QC work?

AdamTheisen:
@kenkehoe the updated qc will not work which is what I try to note in the example and why users should filter their data appropriately ahead of any transformation

kenkehoe:
@AdamTheisen  OK sounds good. Thanks for the example.

PR #770, opened 2023-12-11
Correctly inserting two nan values for 2D plots to ensure minimal streaking in plot.

### LINKED ISSUE #766: Streaking with missing data in 2D plots
### Description

There is a new issue with pcolormesh() causing streaking when there is larger data gaps between data periods. If you look closely at the plot there is a part with some structure at the start and end of the plotted periods. That is the actual data. The extended streaking part is an artifact of pcolormesh(). The issue is that we use add_nan=True to add one NaN value in the middle of the missing data period. pcolormesh() will extend the plot to ½ that window to the inserted NaN. All attempts at changing this behavior with shading keyword (set_shading in .plot() from timeseriesdisplay()) did not fix this issue.
epcwsacrS2 a1 vpt 20231102

### What I Did
If I modify data_utils.add_in_nan() to add two values during the missing data period, one right after last value and one just before next value, the issue is resolved.
Figure_1

I have some code ready to go. It will add one NaN value in the middle of the data period for 1D data, same as normal. But will add a value right after last value and right before next value. With two values added the streaking is removed. With v2.0 coming will need to integrate correctly.

kenkehoe:
There is an issue with plotting multiple data files with a large data gap and 2D data. Inserting on NaN value allows the plot to streak the image. This PR will insert 2 NaN values to stop the streaking.

- [ X] Closes #766 

### DIFF
--- a/act/tests/test_utils.py
+++ b/act/tests/test_utils.py
@@ -86,12 +86,13 @@ def test_add_in_nan():
     time = np.arange('2019-01-01T01:00', '2019-01-01T02:00', dtype='datetime64[m]')
     data = np.random.random((len(time), 25))
 
-    time = np.delete(time, range(3,8))
-    data = np.delete(data, range(3,8), axis=0)
+    time = np.delete(time, range(3, 8))
+    data = np.delete(data, range(3, 8), axis=0)
     time_filled, data_filled = act.utils.add_in_nan(time, data)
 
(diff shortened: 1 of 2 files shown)


mgrover1 (approved):
This looks great! Thanks for the fix here @kenkehoe

zssherman (approved):

AdamTheisen (approved):
@kenkehoe this looks good, we might want to consider adding to the unit tests for this function and that's something I can look at

PR #780, opened 2023-12-21
Catching divide by 0 error before it errors.

kenkehoe:
There is a chance that the division could cause error when the mean is zero. This just catches that instance and handles correctly.

- [ X ] PEP8 Standards or use of linter


zssherman (approved):

zssherman:
Looks good! Merging

PR #795, opened 2024-01-26
Unit conversion warning

### LINKED ISSUE #793: Add warning to change units when fails
Need to add warning to change units method to indicate when a problem occurs, but don't raise an error as many string units are not correct or udunits compliant.

Describe what you were trying to get done.
Tell us what happened, what went wrong, and what you expected to happen.

### What I Did

```
Paste the command(s) you ran and the output.
If there was a crash, please include the traceback here.
```

kenkehoe:
This adds keywords to the convert_units method to notify user when the conversion fail. Because of how the method can loop through all the variables in the Dataset we don't want to print or raise exceptions when a conversion fails. This adds options the user can implement. I tried to use warnings but they didn't work so it's a print statement.

- [ X] Closes #793
- [X ] Tests added

### DIFF
--- a/act/utils/data_utils.py
+++ b/act/utils/data_utils.py
@@ -35,7 +35,8 @@ class ChangeUnits:
         self._ds = ds
 
     def change_units(
-        self, variables=None, desired_unit=None, skip_variables=None, skip_standard=True
+        self, variables=None, desired_unit=None, skip_variables=None, skip_standard=True,
+        verbose=False, raise_error=False
     ):
         """
         Parameters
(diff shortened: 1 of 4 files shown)


act/utils/data_utils.py
```diff
@@ -51,6 +52,13 @@ def change_units(
             Flag indicating the QC variables that will not need changing are
             skipped. Makes the processing faster when processing all variables
             in dataset.
+        verbose : boolean
+            Option to print statement when an attempted conversion fails. Set to False
+            as default because many units strings are not udunits complient and when
+            trying to convert all varialbes of a type of units (eg temperature) the code
```
zssherman, line 58: complient to compliant and varialbes to variables

tests/utils/test_data_utils.py
```diff
@@ -93,10 +95,13 @@ def test_convert_units():
     data = act.utils.data_utils.convert_units(r_data, 'K', 'C')
     assert np.ceil(data[0]) == 12
 
-    try:
+    # try:
+    #     ds.utils.change_units()
```
zssherman, line 99: Can we remove the commented out code?

zssherman:
@kenkehoe Thanks for the PR! I added a few comments

kenkehoe:
@zssherman I removed the old code and submitted a job to run. I assume you can make the failing part of this process more better? I don't think it's something on my end.

zssherman:
@kenkehoe Thanks for the changes! Yeah it was a stalled test. Should be rerunning now

AdamTheisen (approved):

PR #796, opened 2024-01-31
ENH: Bug fix for when cropscape shifts pixel between soybean and wint…

AdamTheisen:
…er wheat


zssherman (approved):
@AdamTheisen Looks good to me! Thanks!

PR #797, opened 2024-02-01
CI: Add linting action that uses pre-commit and ruff.

zssherman:


zssherman:
I need to find a way for it to ignore removing the imports in the try and except blocks

zssherman:
@mgrover1 Is there a way for precommit to fix the 600 or so long line errors? Or do i have to manually go in and fix all 600.... Or is there another way around it?

mgrover1:
@zssherman - you can fix these files using ruff, which is part of the pre-commit hooks

See https://docs.astral.sh/ruff/tutorial/#getting-started

You can run
```bash
ruff check . --fix
```
 to automatically go in and fix those files

zssherman:
@mgrover1 Hmm sadly it doesn't do anything at all from the act directory.

zssherman:
@mgrover1 I think I figured it out, was missing that addition of ruff to the pyproject.yml

mgrover1 (approved):

PR #801, opened 2024-02-14
ENH: New baseline images and updates to plot test tolerances

### LINKED ISSUE #768: Verify that unit tests for plotting are comparing and doing what it's supposed to.
There have always been quirks with the image tests, especially when running on GitHub CI.  Need to perform some baseline testing to ensure it's appropriately comparing plots with the baseline images.  It would also be good to understand what changes to the plots will fall within the threshold for differences and if we should adjust those thresholds.

AdamTheisen:
- [x] Closes #768 


zssherman (commented):
For the multi_skewt plot, there is alot of overlap between plots, I wonder if we need to change a setting for spacing?

AdamTheisen:
@zssherman I was thinking the same thing the more I look at it.  This was one where we had a tolerance of 81 set so it has previously been problematic.

tests/plotting/baseline/test_time_height_scatter2.png
zssherman, line 1: Same for this one

zssherman:
Yeah, we can investigate! I only see the two specific examples, so I can test them out and check the module.

zssherman (approved):

PR #803, opened 2024-02-23
ENH: Bug fix for set_xrng

AdamTheisen:


AdamTheisen:
Looks like we might have had a dependency update that's causing other code to fail.

zssherman (approved):

zssherman:
Hmm yeah, I can take a look! Yeah might just need to add a new parameter when we decode times

PR #804, opened 2024-02-23
FIX: Fix for new xarray check for dtype.

zssherman:
Trying to understand the encoding was a learning experience to say the least. The input encoding is 'float64' so I set it to that. 'datetime64[ns]' does not work.

### DIFF
--- a/act/io/arm.py
+++ b/act/io/arm.py
@@ -767,6 +767,9 @@ class WriteDataset:
                 ]
             )
 
+        if 'time_bounds' in encoding.keys():
+            encoding['time_bounds']['dtype'] = 'float64'
+
         if hasattr(write_ds, 'time_bounds') and not write_ds.time.encoding:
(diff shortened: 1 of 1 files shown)


AdamTheisen:
@kenkehoe can we always assume that time bounds are always going to be floats?

zssherman:
Yeah if it does change, I can try to find a way to automate it.

kenkehoe:
I believe time_bouds by ARM standards will always be a float (type float or double). It should be the number of seconds to add and subtract to define the bound. It should match the same type as the written time coordinate variable. Since Xarray converts time upon reading to datetime64 but not time_bounds it should stay a float.

mgrover1 (approved):

zssherman:
@AdamTheisen Good to merge?

AdamTheisen (approved):

PR #811, opened 2024-03-25
ADD: calculate_percentage function.

zssherman:
This function allows the user (ex: aerosol data) to provide fields and calculate the percentage of that field in a volume. The code also checks for invalid numbers if the user provides a threshold. User can also provided a specific time, a range of times, or get an mean of the whole dataset.

This function will eventually be used for the PieChart function I'm working on.

### DIFF
--- a/act/tests/__init__.py
+++ b/act/tests/__init__.py
@@ -60,6 +60,7 @@ __getattr__, __dir__, __all__ = lazy.attach(
             'EXAMPLE_ENA_MET',
             'EXAMPLE_CCN',
             'EXAMPLE_OLD_QC',
+            'EXAMPLE_AOSACSM',
         ]
     },
 )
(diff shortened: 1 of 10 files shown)


zssherman:
I'm not sure why the hashes aren't matching...

zssherman:
Rerunning one more time with the PR now for arm-test-data

zssherman:
@AdamTheisen Yeah I can do that! It's a one liner that I can added in the example. Once the piechart function is actually added, i can swap out the one line in the example with the new function

zssherman:
@AdamTheisen Example has been added.

zssherman:
Once this is approved, we can merge and I';; rerun Bobby's PR

AdamTheisen (approved):
@zssherman looks good!

PR #812, opened 2024-03-27
FIX: Small error in Doppler lidar code with undefined var_name

rcjackson:
The VAD code would not work if only signal_to_noise_ratio was defined and intensity was set to None due to an undefined variable name. This remedies this by pointing the var_name to be the snr_name which is only used to get the height_name.


zssherman (approved):

zssherman:
Failure is unrelated and is fixed in my PR

AdamTheisen (approved):

PR #822, opened 2024-04-11
FIX: Many warning fixes for ACT.

zssherman:
These fixes include datetime utc changes, missing images, xarray.dims to sizes changes and more.

Note I did a PR for icartt as well to fix there warnings, which will then suppress ours and will be doing one for PySP2 as well.


zssherman:
For more on timezone fun, see here:
https://github.com/Unidata/MetPy/issues/3298

AdamTheisen (commented):
@zssherman  are those barb baseline plots what we're producing now?  Seems like there's not a lot there... No actual barbs.  Thoughts?

zssherman:
@AdamTheisen Originally the plots u_v 3, 4 and5 were not producing plots and had a warning of future errors because of it, so I added the plot comparison plot. In terms of the small amounts of data in them, compared to u_v_1 I'm seeing the use of a real dataset, were the rest use a small fake ds, which could be the reason for that? Also the bins are set low it looks like. So i'm not 100% if because they are test ds that there is barely any data because its np.linspace fake data? But that was my assumption when I saw the plots.

zssherman:
@AdamTheisen Sounds good!

PR #828, opened 2024-05-01
ADD: Adding a piechart function to distribution display module.

### LINKED ISSUE #550: Add a PieChartDisplay to Support the Aerosol Community
Pie charts are helpful when working with aerosol data. We should have this functionality in ACT.

zssherman:
This new function uses act.utils.calculate_percentages to create a pie chart of percentages of user defined fields of a dataset.

- [ ] Closes #550
- [ ] Tests added

piechart


act/plotting/distributiondisplay.py
```diff
@@ -863,3 +863,83 @@ def plot_violin(
...
+        set_title : str
+            The title of the plot.
+        autopct : str
+            Format string for the percentages. Default is float with one
+            decimal place. If this parameter is set to None, no percentage
+            string values are displayed.
+        percent_kwargs : dict
+            Dictionary of parameters to be used in the act.utils.calculate_percentages
+            function. None will use default parameters.
+
```
AdamTheisen, line 898: @zssherman do we need to list kwargs down here in the parameters list?

zssherman: @zssherman My view on it would be to keep it, just because we are using two kwargs, one for ax.pie and the other for calculate percentages

AdamTheisen: Sorry, I was just referring to the documentation and just adding 2 lines to list the kwargs as a parameter.  I didn't see it listed in the documentation.  

@mgrover1 you're noting that for the percent_kwargs right and not for the kwargs passed to the matplotlib pie function right?

zssherman: @AdamTheisen There is a mention below that shows it goes to ax.pie. And yeah Max and I discussed he was thinking for verbose, to add the parameter from calculated_percentages as there is not many and it easier to read and access.

zssherman: @AdamTheisen I added in the doc the kwargs part to be consistent with the rest of the act docs. Also did the changes @mgrover1 suggested.

AdamTheisen (approved):

PR #832, opened 2024-05-20
Getting correct length of MPG movie.

kenkehoe:
There is an issue guessing the duration of the MPG movie when converting to another file type. If the time step of the movie is more than 0.1 then there can be multiple images that are the same. Checking multiple images being the same is how to determine the end of the movie. This fix allows some fudge factor by checking for multiple repeating images to signify the end.


zssherman (approved):
Looks like a reasonable fix to me! I've never heard of the deque module before, learned something new.

PR #835, opened 2024-05-22
Fix HYSPLIT reader to adjust for an arbitrary number of trajectories.

rcjackson:
The HYSPLIT reader was assuming that the data started at line 13, which was breaking the reader when using other test files generated by students at the ARM Summer School. This now fixes that issue, making the HYSPLIT reader able to read files with an arbitrary number of trajectories.

### DIFF
--- a/act/io/hysplit.py
+++ b/act/io/hysplit.py
@@ -35,8 +35,12 @@ def read_hysplit(filename, base_year=2000):
             num_lines += 1
             grid_names.append(data[0])
             grid_times.append(
-                datetime(year=(int(data[1])+ base_year),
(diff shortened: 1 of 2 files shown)


zssherman (approved):

mgrover1 (approved):

AdamTheisen (commented):
Do we want to have skiprows as an argument in case we have other variations of the file?

AdamTheisen:
@zssherman @rcjackson what are your thoughts on adding skiprows as an argument in case we run into different flavors of this in the future?

rcjackson:
Went ahead and made the change @AdamTheisen.

AdamTheisen (approved):
Thanks @rcjackson !

PR #855, opened 2024-09-12
ADD: MPLNET Discovery Module

jrobrien91:
- [X] Tests added

Notes: 
- MPLNET NETCDF4 files (at least Level 1) have some CFWarnings and fail our `read_arm_netcdf` I/O module. This is primarily due to the altitude dimension not containing a corresponding 1D coordinate within the files. 
- Will work on I/O function to correct these CFWarnings within the download files in a followup PR. 
- Module was named `nasa.py` as I anticipate eventually including discovery modules for NASA AERONET and PANDORA APIs as well.

### DIFF
--- a/act/discovery/__init__.py
+++ b/act/discovery/__init__.py
@@ -8,7 +8,7 @@ import lazy_loader as lazy
 
 __getattr__, __dir__, __all__ = lazy.attach(
     __name__,
-    submodules=['arm', 'asos', 'airnow', 'cropscape', 'improve', 'noaapsl', 'neon', 'surfrad'],
+    submodules=['arm', 'asos', 'airnow', 'cropscape', 'improve', 'noaapsl', 'neon', 'surfrad', 'nasa'],
     submod_attrs={
         'arm': ['download_arm_data', 'get_arm_doi'],
         'asos': ['get_asos_data'],
@@ -18,5 +18,6 @@ __getattr__, __dir__, __all__ = lazy.attach(
         'noaapsl': ['download_noaa_psl_data'],
         'neon': ['get_neon_site_products', 'get_neon_product_avail', 'download_neon_data'],
         'surfrad': ['download_surfrad_data'],
+        'nasa' : ["download_mplnet_data", "get_mplnet_meta"]
     },
 )
--- a/act/discovery/__init__.py
+++ b/act/discovery/__init__.py
@@ -8,7 +8,17 @@ import lazy_loader as lazy
 
 __getattr__, __dir__, __all__ = lazy.attach(
     __name__,
-    submodules=['arm', 'asos', 'airnow', 'cropscape', 'improve', 'noaapsl', 'neon', 'surfrad', 'nasa'],
+    submodules=[
+        'arm',
+        'asos',
+        'airnow',
+        'cropscape',
+        'improve',
+        'noaapsl',
+        'neon',
+        'surfrad',
+        'nasa',
+    ],
     submod_attrs={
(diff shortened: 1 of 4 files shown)


act/discovery/__init__.py
```diff
@@ -18,5 +18,6 @@
         'noaapsl': ['download_noaa_psl_data'],
         'neon': ['get_neon_site_products', 'get_neon_product_avail', 'download_neon_data'],
         'surfrad': ['download_surfrad_data'],
+        'nasa' : ["download_mplnet_data", "get_mplnet_meta"]
```
AdamTheisen, line 21: I believe there are some linting errors here.  Did you run flake8 on the code at all?

examples/discovery/download_mplnet.py
AdamTheisen, line 1: We might need to rename to plot_mplnet.py so that the documentation picks it up.  It also might be beneficial to show how to read those files in and plot it.

act/discovery/__init__.py
jrobrien91: I did and came back fine using `python -m flake8 --max-line-length=127 --ignore=F401,E402,W504,W605,F403`

Pre-commit and ruff must be stricter linters

jrobrien91: I also installed `pre-commit` after I had made commits, when the repo mentioned I should check the updated contributors guide prior to this PR. So I am sure that's how they were missed

examples/discovery/download_mplnet.py
AdamTheisen: @jrobrien91 I tried running the example but I get a "Failed MPLNET Download" error

jrobrien91: @AdamTheisen thank you for catching that. I had a processing level leftover from testing in there. It should be level=1. I submitted a quick fix

AdamTheisen:
@jrobrien91 will try to get to reviewing soon.  In reviews and meetings all day/week.

AdamTheisen (approved):
Looks to be working!  Just looking ahead, I tried to read the file in with with just xarray and it is problematic so definitely not a straightforward way.

xarray.core.variable.MissingDimensionsError: 'altitude' has more than 1-dimension and the same name as one of its dimensions ('time', 'altitude'). xarray disallows such variables because they conflict with the coordinates used to label dimensions.

jrobrien91:
> Looks to be working! Just looking ahead, I tried to read the file in with with just xarray and it is problematic so definitely not a straightforward way.
> 
> xarray.core.variable.MissingDimensionsError: 'altitude' has more than 1-dimension and the same name as one of its dimensions ('time', 'altitude'). xarray disallows such variables because they conflict with the coordinates used to label dimensions.

@AdamTheisen So far I have a fix for Altitude and am able to read these into xarray, but these are some funky files. Over half the dimensions do not have assigned coordinates which is breaking `open_mfdataset`. Working on an xarray friendly I/O now with @zssherman

AdamTheisen:
@jrobrien91 awesome!  I would defer to you if you want to wait to have it in this PR or another one.  This looks good to move forward with on my end but will defer to @zssherman 

I'm not sure if it would be helpful but the ARM MMCR files are problematic with xarray as well and we had to do some base NetCDF workarounds.  You can check out the code at the following link if that helps: https://github.com/ARM-DOE/ACT/blob/main/act/io/arm.py#L817

zssherman:
@AdamTheisen I would be fine with merging this in! We can then work on updating the example and having the actual IO separately.

PR #857, opened 2024-09-25
Persistence test ignore range

kenkehoe:
This PR updates the persistence test to have a new keyword to allow ignoring the results of the test when the data values fall within a range of values. There are known times when values will fail the persistence test but there is not issue. e.g. weather station during a rain event has RH at 100%.

- [X ] Tests added

### DIFF
--- a/act/qc/qctests.py
+++ b/act/qc/qctests.py
@@ -948,6 +948,7 @@ class QCTests:
         test_number=None,
         flag_value=False,
         prepend_text=None,
+        ignore_range=None,
     ):
         """
         Method to perform a persistence test over 1-D data..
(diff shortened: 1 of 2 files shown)


kenkehoe:
OK I will look into this and find a solution for the hardware differences.

zssherman:
Hmm, the windows test failing is interesting, wonder if there was some sort of change to actions for windows builds?

zssherman:
Error is unrelated to the PR, windows builds are broken for Py-ART as well.

zssherman:
Once the non windows tests pass, I will work on CI fixes in another PR.

PR #888, opened 2024-12-11
FIX: `read_arm_mmcr` fails when using a read-only input file

### LINKED ISSUE #887: ACT fails to run `read_arm_mmcr` when using read-only files as input
* ACT version: 2.1.4
* Python version: 3.11.5
* Operating System: Red Hat v7.9

### Description

ACT fails to run `read_arm_mmcr` when using read-only files as input

### What I Did

Upon running the following code:

```
import act
filename = './sgpmmcrmomC1.b1.20030927.000000.cdf'
mmcr_file = act.io.read_arm_mmcr([filename])
mmcr_file
```

I receive the following error:

```
OSError                                   Traceback (most recent call last)
Cell In[1], line 3
      1 import act
      2 filename = './sgpmmcrmomC1.b1.20030927.000000.cdf'
----> 3 mmcr_file = act.io.read_arm_mmcr([filename])
      5 mmcr_file

File ~/git_dev/ACT/act/io/arm.py:844, in read_arm_mmcr(filenames)
    842 multi_ds = []
    843 for f in filenames:
--> 844     nc = Dataset(f, 'a')
    845     # Change heights name to range to read appropriately to xarray
    846     if 'heights' in nc.dimensions:

File src/netCDF4/_netCDF4.pyx:2521, in netCDF4._netCDF4.Dataset.__init__()

File src/netCDF4/_netCDF4.pyx:2158, in netCDF4._netCDF4._ensure_nc_success()

OSError: [Errno 30] Read-only file system: './sgpmmcrmomC1.b1.20030927.000000.cdf'
```

isilber:
- [x ] Closes #887 
- [X ] PEP8 Standards or use of linter

### DIFF
--- a/act/io/arm.py
+++ b/act/io/arm.py
@@ -841,12 +841,12 @@ def read_arm_mmcr(filenames):
     # read it in with xarray
     multi_ds = []
     for f in filenames:
-        nc = Dataset(f, 'r')
+        nc = Dataset(f, 'a', diskless=True)
+        # Change heights name to range to read appropriately to xarray
+        if 'heights' in nc.dimensions:
+            nc.renameDimension('heights', 'range')
         if nc is not None:
             ds = xr.open_dataset(xr.backends.NetCDF4DataStore(nc))
-            # Change heights name to range to read appropriately to xarray
-            if 'heights' in ds.dims:
-                ds = ds.rename_dims({"heights": "range"})
             multi_ds.append(ds)
     # Concatenate datasets together
     if len(multi_ds) > 1:
--- a/act/io/arm.py
+++ b/act/io/arm.py
@@ -841,12 +841,11 @@ def read_arm_mmcr(filenames):
     # read it in with xarray
     multi_ds = []
     for f in filenames:
-        nc = Dataset(f, 'a')
-        # Change heights name to range to read appropriately to xarray
-        if 'heights' in nc.dimensions:
-            nc.renameDimension('heights', 'range')
+        nc = Dataset(f, 'r')
         if nc is not None:
             ds = xr.open_dataset(xr.backends.NetCDF4DataStore(nc))
+            # Change heights name to range to read appropriately to xarray
+            ds = ds.rename_dims({"heights": "range"})
             multi_ds.append(ds)
     # Concatenate datasets together
     if len(multi_ds) > 1:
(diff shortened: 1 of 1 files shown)


act/io/arm.py
```diff
@@ -841,12 +841,12 @@ def read_arm_mmcr(filenames):
     # read it in with xarray
     multi_ds = []
     for f in filenames:
-        nc = Dataset(f, 'a')
-        # Change heights name to range to read appropriately to xarray
-        if 'heights' in nc.dimensions:
-            nc.renameDimension('heights', 'range')
+        nc = Dataset(f, 'r')
         if nc is not None:
```
zssherman, line 845: Is there a reason for this if check?

act/io/arm.py
```diff
@@ -841,12 +841,12 @@ def read_arm_mmcr(filenames):
...
             ds = xr.open_dataset(xr.backends.NetCDF4DataStore(nc))
+            # Change heights name to range to read appropriately to xarray
+            if 'heights' in ds.dims:
+                ds = ds.rename_dims({"heights": "range"})
```
zssherman, line 849: We are using heights later on for some calculations, this might break

act/io/arm.py
isilber: @zssherman, I guess this was a legacy thing, so I didn't modify it.
Ideally, I'd rewrite that method using solely xarray and making it much shorter, but I do not have the time to do it at the moment, so in many aspects, this is more of a patch.

act/io/arm.py
isilber: @zssherman Well, it doesn't, and I've tested it. 
As it stands, ARM currently holds only the b1 MMCR files, which have that dimension. The issue was with read-only files. Other than that, everything remained the same.
You can try the few lines of code in the issue I opened and see that it works.

zssherman: Yeah i was curious on why we are checking for nans later on etc for height and if that would break if we had range, but I see we check for range as well. Thanks for the clarification.

zssherman:
Tests failing is unrelated to this PR

AdamTheisen:
@isilber  I was running into errors when using this change on my local machine but I've tested this out on read only files and seems to work.

AdamTheisen (approved):
I tested this fix out on ARM's dev system and was able to read from /data/archive.  I think we can go ahead and merge this in @zssherman.  Thanks all!

PR #916, opened 2025-04-02
ADD: Adding ameriflux discovery module. This module is based on the R code amerifluxr

zssherman:
- [ ] Tests added


zssherman:
Note, I still need to add tests and examples

zssherman:
@AdamTheisen Should we leave the default to agree_policy False? So users read the policy first? I was debating how to approach that...

AdamTheisen:
@zssherman I didn't get a chance to review today but blocking my calendar Wed afternoon

zssherman:
@AdamTheisen no worries!

act/discovery/ameriflux.py
```diff
@@ -0,0 +1,265 @@
...
+         "other_research": "Research - Other",
+         "education": "Education (Teacher or Student)",
+         "other": "Other"
+    description : str
+        Brief description of intended use. This will be recorded in the data download log and emailed to site’s PI.
+    out_dir : str
+        The output directory for the data. Set to None to make a folder in the
+        current working directory with the same name as *datastream* to place
+        the files in.
+    is_test : bool
```
AdamTheisen, line 68: Looking at the AmeriFlux recommendations on not making this too easy to set to true.  I'm wondering if we need to only allow this once we the username/password we will use for Git Testing.  Thoughts?

zssherman: Ah good point! Yep I'll put a change in

act/discovery/ameriflux.py
```diff
@@ -0,0 +1,265 @@
...
+            "other": "Other",
+        }.get(intended_use)
+
+    if not intended_use_extended(intended_use):
+        raise ValueError("Invalid intended_use input...")
+
+    # Check if out_dir is reachable
+    if out_dir is None:
+        out_dir = os.path.abspath(os.path.expanduser("~"))
+    if not os.path.exists(out_dir):
```
AdamTheisen, line 141: Instead of an error, would we want to make the directory?

AdamTheisen: Also, it doesn't seem to recognize when I try and pass ./data/ameriflux/

act/discovery/ameriflux.py
```diff
@@ -0,0 +1,265 @@
...
+        raise ValueError("out_dir not valid...")
+
+    # Prompt for data policy agreement
+    if data_policy == "CCBY4.0":
+        warnings.warn(
+            "Data use guidelines for AmeriFlux CC-BY-4.0 Data Policy:\n"
```
AdamTheisen, line 147: May want to add newlines at the start and end of this warning.

zssherman:
@AdamTheisen I did changes to address your comments. Let me know if the recent commit looks good for all your comments. For the is test I add a kwargs, but once we have the seceret email and username added, ill add one more check to the is test

zssherman:
@AdamTheisen Added the new reader, ready for review when folks have time. Working on the tests still though

act/discovery/ameriflux.py
zssherman: This should be fixed now

act/discovery/ameriflux.py
zssherman: I believe I changed the lines as requested

act/discovery/ameriflux.py
zssherman: I hid the parameter in a kwargs with no documentation etc

zssherman:
New reader has been added, working on tests

PR #918, opened 2025-04-15
ENH: Updating to work with moviepy v2 and above.

AdamTheisen:
- [ ] Closes #xxxx
- [ ] Tests added

### DIFF
--- a/act/utils/io_utils.py
+++ b/act/utils/io_utils.py
@@ -355,9 +355,9 @@ def generate_movie(images, write_filename=None, fps=10, **kwargs):
 
                         duration += 0.1
 
-                    clip = clip.set_start(0)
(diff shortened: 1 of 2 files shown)


zssherman:
@AdamTheisen I think we just need to update the unit test file as well. I think that's the cause of the failing tests

mgrover1:
> @AdamTheisen I think we just need to update the unit test file as well. I think that's the cause of the failing tests

I agree with @zssherman here! Updating that file should fix the failing CI here.

AdamTheisen:
@zssherman last test and we should be good to go!

mgrover1 (approved):

zssherman (approved):

PR #926, opened 2025-05-15
Removing duration for mpg movies calculation.

kenkehoe:
Removing duration calculation for mpg movies calculation. moviepy 2.* appears to have fixed this. Added duration keyword for movie generation to allow setting in case moviepy gets it wrong. Added additional check for moviepy version since that is now required for duration setting.


zssherman (approved):
Looks reasonable to me, thanks for adding this changes!

PR #928, opened 2025-05-27
FIX: Update extract function to new standards.

zssherman:
Updating the tar function so that it follows the changes stated for python 3.14. See docs here:

https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extractall

@AdamTheisen @mgrover1 If you both could review and see if this is fine. 'data' according to the docs looks to be the safest.

### DIFF
--- a/act/utils/io_utils.py
+++ b/act/utils/io_utils.py
@@ -132,7 +132,7 @@ def unpack_tar(
     for tar_file in tar_files:
         try:
             tar = tarfile.open(tar_file)
-            tar.extractall(path=out_dir)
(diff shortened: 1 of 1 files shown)


act/utils/io_utils.py
```diff
@@ -132,7 +132,7 @@ def unpack_tar(
     for tar_file in tar_files:
         try:
             tar = tarfile.open(tar_file)
-            tar.extractall(path=out_dir)
+            tar.extractall(path=out_dir, filter='data')
```
AdamTheisen, line 135: Should we add this in as a keyword option as well in case users run into some issues?

PR #938, opened 2025-06-09
FIX: os path join was adding windows path for url, which breaks.

zssherman:
Remove os path join, not needed here


zssherman:
@mgrover1 This should finally fix the issue, fingers crossed.

mgrover1 (approved):

Issue #943, opened 2025-07-02
Inconsistent Behavior Due to GET Request


shahchiragh:
**Describe the bug**
Function `arm_site_location_search` in https://github.com/ARM-DOE/ACT/blob/main/act/utils/data_utils.py#L1254 currently uses a GET request to perform ES search action that can lead to inconsistent behavior, particularly when parameters change or when intermediate proxies/middleware cache GET requests inappropriately. This may result in unreliable execution across environments.

**To Reproduce**
1. Trigger `arm_site_location_search` function that uses the GET request.
2. Change the `site_code` parameter to some other site e.g. nsa
3. With different parameters, the response should be different however, its giving same response.

**Expected behavior**
The function should behave consistently regardless of parameter or environmental caching. Switching to a POST request should eliminate these edge case failures by placing parameters in the request body and avoiding caching issues.

**Screenshots**
Current scenario gives same response with different parameters: 
![Image](https://github.com/user-attachments/assets/47122ef3-64e4-4910-bc5a-79037d177a1f)

Scenario after changing the request type:
![Image](https://github.com/user-attachments/assets/3613c8b8-52bf-430f-b20a-4251488e5e60)

**Desktop**
 - OS: Applicable to all OS
 - Browser: Applicable to all Browsers
 - Version: Applicable to all Versions

**Additional context**
Opening a PR for switching from GET to POST to ensure robust handling of metadata function calls.

zssherman:
@shahchiragh Thank you for opening the issue for tracking and for working on a fix!

AdamTheisen:
#944 looks to have resolved this issue

PR #944, opened 2025-07-02
Fix: Update request_type to POST for arm site location search with elastic search

shahchiragh:
- [x] Closes https://github.com/ARM-DOE/ACT/issues/943


zssherman:
@shahchiragh Everything is passing! Thanks again for making those changes! Merging.

PR #1002, opened 2026-04-22
FIX: Update timeseries testing errors to use pytest over numpy.

zssherman:


zssherman:
@AdamTheisen This should test only on linux for the nasa mplnet and also should fix most of the dep warnings for numpy.raises

AdamTheisen (approved):

PR #1023, opened 2026-06-03
ENH: Update publishing method.

zssherman:


zssherman:
@rcjackson Figure to ask if possible for a review since you have dealt with the trusted publishing.

PR #1030, opened 2026-07-01
ADD: reader for mfrsr7nchaod files

### LINKED ISSUE #1018: cftime and open_mfdataset conflict
**Describe the bug**
During the 2026 ARM Summer School, when attempting to open MFRSRAOD data via open_mfdataset, users received errors associated with the additional coordinate fields denoting the calibration times, particularly non-monotonic global indexes along these additional dimensions. 

From the user: 
```
The mfrsr7nchaod1michM1.c1 datastream contains the normal ARM usage of base_time, time_offset, and time, each with units specified as “second since …”.  But in addition, it also contains two extra fixed-length coordinates fields “Io_interquartile_time” and “Io_gauss_time” that specify the times of calibration values.  Here is how they are defined:
 
float Io_interquartile_time(Io_interquartile_time=72);
  :long_name = "Time for interquartile Io sample times";
  :units = "seconds since 1970-01-01 00:00:00";
 
float Io_gauss_time(Io_gauss_time=61);
  :long_name = "Time for Ios after gaussian filter";
  :units = "seconds since 1970-01-01 00:00:00";
```

Issue appears to be within 200+ ARM datastreams that follow this cftime convention

**To Reproduce**
ds = xr.open_mfdataset(mfrsr7nchaod1michM1.c1_files[-3:])

**Expected behavior**
ds = xr.open_mfdataset(files[-3:])
ValueError: Resulting object does not have monotonic global indexes along dimension Io_gauss_time

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**
Will forward the email chain to maintainers

cgodine:
Adding reader for MFRSRAOD files. Will read in a list of filenames, open xarray Datasets of each file, and merge all Datasets along time dimension. Coordinate variable `Io_gauss_time` will have the union of date values from all read files. 

Added `mfrsr7nchaod1mich` files from SGP C1 to `arm-test-data` in order to complete testing.
 

- [X] Closes #1018 
- [X] Tests added

### DIFF
--- a/act/io/__init__.py
+++ b/act/io/__init__.py
@@ -27,6 +27,7 @@ __getattr__, __dir__, __all__ = lazy.attach(
             'read_arm_netcdf',
             'check_if_tar_gz_file',
             'read_arm_mmcr',
+            'read_arm_mfrsr7nchaod',
         ],
         'ameriflux': ['convert_to_ameriflux', 'read_ameriflux'],
         'text': ['read_csv'],
--- a/act/io/arm.py
+++ b/act/io/arm.py
@@ -908,3 +908,88 @@ def read_arm_mmcr(filenames):
             ds[new_var_name] = da
 
     return ds
+
+
+def read_arm_mfrsr7nchaod(filenames):
+    """
+
+    Reads in mfrsr7nchaod files and merges along time dimension. MFRSR7NCHAOD files have
+    coordinate dimensions Io_interquartile_time and Io_gauss_time that cannot be concatenated
+    in xarray.open_mfdataset() due to overlapping values within adjacent files.
+
+    Parameters
+    ----------
+    filenames : str, pathlib.PosixPath, or list of str
+        Name(s) of fullpath(s) to read.
+
+    Returns
+    -------
+    ds : xarray.Dataset or None
+        ACT Xarray Dataset. If no files found will return None.
+    """
+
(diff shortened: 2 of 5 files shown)


zssherman:
@cgodine Thanks for the PR! Are these files netcdf? What I'm debating is, if possible, if we would want to integrate this with the current arm netcdf reader or if its possible, or if a separate reader is fine. @AdamTheisen thoughts? Maybe the same process as the current reader, but then a check for these types of files that feeds them through code like we see in this PR?

cgodine:
> @cgodine Thanks for the PR! Are these files netcdf? What I'm debating is, if possible, if we would want to integrate this with the current arm netcdf reader or if its possible, or if a separate reader is fine. @AdamTheisen thoughts?

Yes, they are netcdf (ARM c1-level VAP files). I had thought about integrating with the read_arm_netcdf function, but I believe this is the only ARM data with the rolling window coordinate dimension causing the concatenation issue. I can definitely refactor if that is a more desirable outcome though!

AdamTheisen:
@zssherman I would agree with that.  I do wonder if we can do something in the main reader to account for this or make this more generic so it's not just for this 1 specific datastream.

AdamTheisen:
Closing pull request and will work towards a more generic solution

PR #1031, opened 2026-07-06
FIX: add exception handling for when CAPE and CIN cannot be calculated

cgodine:

When attempting to calculate stability indices for a sonde profile with an LCL above the top of the temperature profile, an `IndexError` is raised because the environmental profile never intersects with the LCL (see below image for an example).

dstsondewnpnM1 b1 enhanced_skewt 173000 20260705

This means that `metpy.calc` cannot interpolate a parcel profile and thus cannot calculate CAPE or CIN values. In this case, the CAPE and CIN values will be set to `np.nan`.


zssherman:
Restarting test, now that neon discovery is fixed

zssherman:
@cgodine As a heads up, looks like your code changes from the other PR are in this PR as well

cgodine:
Thanks for the heads up @zssherman, I reverted those changes.

AdamTheisen (approved):

zssherman:
@cgodine The only failure test wise im seeing is a difference still in this test:
FAILED tests/retrievals/test_sonde.py::test_get_stability_indices - AssertionError: 
Arrays are not almost equal to 2 decimals
 ACTUAL: <xarray.DataArray 'surface_based_cape' ()> Size: 8B
array(0.9845778)
Attributes:...
 DESIRED: 0.96

PR #1060, opened 2026-09-11
ENH: Add QC- and bounds-aware data transformations

### LINKED ISSUE #703: QC-aware transformations
xarray provides powerful [resample](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.resample.html) and [groupby](https://docs.xarray.dev/en/stable/getting-started-guide/quick-overview.html#groupby) methods for transforming data onto a target coordinate grid, but it has no concept of QC variables so any transformations applied (e.g., mean, nearest) are performed naively and could result in the use of data that has been flagged as bad.

I think ACT would be a great place to host extensions to xarray's methods that do account for QC values in transformations. The proposed interface would be a series of methods that mirror the transformation types offered by the ARM Data Integrator (ADI) made available to ARM users in the PCM Interface:

* nearest neighbor
* bilinear interpolation
* bin averaging
* auto (picks between interpolation and averaging based on bin size) -- optional

The ADI library makes a few key decisions for QC-aware transformations that I think should be mirrored here:
* data values QC'd as bad are excluded from consideration in the transformation
  * for averaging this is equivalent to marking these as NaN
  * for nearest neighbor(interpolation) the nearest non-bad point(s) are used
* for averaging, if >*threshold* % of values in a bin are bad, then the output value is set to missing
  * also if >*other threshold* are indeterminate, output value is also set to missing
  * there are reasonable defaults set for each (I *think* these are 50% and 80%, respectively)
  * PCM/ADI goes a step further and lets you customize this for each variable, but I'm not sure if the functionality/complexity trade-off is worth it for ACT
* optionally, an output summary QC variable is generated for each QC'd input variable

I think this could be implemented as a method applied to an xarray `DatasetResample` / `DatasetGroupBy` object returned by `ds.resample` / `ds.groupby`, e.g.,:

```python
# Proposed API

import act
import xarray as xr


ds = act.io.armfiles.read_netcdf(...)

ds.resample(time="30min").apply(
    act.qc.transform.NearestNeighbor(tolerance="15min")
)

ds.resample(time="30min").apply(
    act.qc.transform.Interpolate(method="linear")
)

ds.groupby("time.hour").apply(
    act.qc.transform.BinAverage(
        bad_threshold=0.5,
        indeterminate_threshold=0.8,
        add_transform_qc=False,  # maybe also a roll-up QC option like PCM (4 bits instead of 10+)
    ),
)
```

The transform functions/classes (`NearestNeighbor`, `Interpolate`, `BinAverage`) should take and return xarray `Dataset` objects. The input passed by the `apply` method contains all the points in the given bin and the output is expected to be a 0-coord Dataset with scalar values for each data variable (metadata included).

I'm totally open to any changes/feedback. This could probably use several iterations of revisions to make it easier for users. Let me know what you think!

maxwelllevin:
## What changed

- Added a new `act.transform` submodule providing QC- and bounds-aware:
  - Bin averaging
  - Linear interpolation
  - Nearest-neighbor subsampling
- Implemented the transformations based on ARM’s [libtrans](https://github.com/ARM-DOE/ADI/tree/master/packages/libtrans) library.
- Added equivalent transformation methods through the `.transform` xarray dataset accessor.
- Added support for calling the transformations directly as public functions.
- Added propagation and handling of:
  - Quality-control variables
  - Quality-control bit definitions
  - Coordinate bounds
  - `datetime64` and `cftime` coordinates
- Added tests, examples, and user/API documentation.

## Why

Xarray’s built-in transformation methods do not natively account for quality control variables. In addition, `xarray.resample().mean()` considers only coordinate values and not coordinate bounds. It therefore assumes that each measurement belongs entirely to a single output bin, which can produce incorrect results when measurements span bin boundaries.

These transformations address both issues while following the behavior of ARM’s existing `libtrans` implementation.

## Dependency considerations

The implementation uses `numba` to JIT-compile the loop-based transformation logic. The code falls back to regular python execution without `numba`, which may be significantly slower. Making `numba` an optional dependency could be considered.

## Issues closed

- Closes #703 

## Validation

- [x] Ran pre-commit/ruff successfully.
- [x] Added and ran tests covering:
  - The public transform API
  - `.transform` dataset accessor behavior
  - Datetime and `cftime` coordinate/bounds handling
  - ARM-compatible QC propagation and bit definitions
  - Numba-enabled and fallback execution
- [x] Added and reviewed documentation examples.
- [x] Confirmed transformation results are consistent with ARM’s C implementations.

## AI usage & manual verification

- AI used: yes
- Model/tooling: GPT-5.6 Luna with Claude Code
- AI assistance included coding, debugging, testing, documentation, and example generation.
- I manually reviewed the changes, ran the relevant tests and examples, inspected the generated documentation, and verified the transformation output against ARM’s C implementations.

## Screenshots

<details><summary>Details</summary>

user guide 1

examples overview

example data array transform

example dataset transform

</details>


act/transform/driver.py
```diff
@@ -0,0 +1,887 @@
...
+                if target_ds is not None and dim in target_ds.coords:
+                    ob = _extract_bounds(target_ds[dim], None, target_ds)
+                    if ob is not None:
+                        var_kwargs['output_bounds'] = ob
+
+        out, out_qc = fn(da, target_da, dim, qc=qc_da, qc_mask=qc_mask, **var_kwargs)
+        result_vars[name] = out
+        result_vars[qc_name] = out_qc
+
+    return xr.Dataset(result_vars, attrs=ds.attrs)
```
AdamTheisen, line 811: For the output from this function, time has an attribute of "bounds: time_bounds" but time_bounds is not in the returned dataset.  Do we have a metadata mismatch here or should we be including time_bounds in the returned dataset?

```
Coordinates:
  * time     (time) datetime64[ns] 384B 2026-08-27T00:15:00 ... 2026-08-27T23...
Attributes:
    long_name:      Time offset from midnight
    standard_name:  time
    bounds:         time_bounds


time bounds attribute: time_bounds
time_bounds exists: False
```

act/transform/driver.py
```diff
@@ -0,0 +1,887 @@
...
+    out_2d = np.full((nt, n_other), output_missing_value, dtype=np.float64)
+    qc_out_2d = np.zeros((nt, n_other), dtype=np.int32)
+
+    if transform == 'bin_average':
+        in_start, in_end = _get_bounds(input_coord, input_bounds)
+        out_start, out_end = _get_bounds(output_coord, output_bounds)
+        for col in range(n_other):
+            stdev = np.full(nt, output_missing_value)
+            coverage = np.full(nt, output_missing_value)
+            _bin_average_1d(
```
AdamTheisen, line 268: _bin_average_1d() returns a status code (0 success, negative on error) but the return value is discarded here.  For example, incompatible input/target ordering returns -5; because out_2d is initialized to the missing value and qc_out_2d to zero, this currently returns all-missing data with QC=0 rather than surfacing the failure. I reproduced this with real sgpmetE13.b1 data using a descending target. Could we capture the status and raise on negative values?

output data initialized to -9999
QC initialized to 0
        ↓
_bin_average_1d() sees incompatible ordering
        ↓
kernel returns -5 immediately
        ↓
driver ignores -5
        ↓
output remains -9999
QC remains 0
        ↓
function returns normally

act/transform/driver.py
```diff
@@ -0,0 +1,887 @@
...
+            continue
+
+        # Skip variables that are themselves QC companions
+        if isinstance(name, str) and name.startswith(qc_prefix):
+            continue
+
+        qc_name = f'{qc_prefix}{name}'
+        qc_da = ds[qc_name] if qc_name in ds else None
+
+        var_transform = per_var_transform.get(name, transform)
```
AdamTheisen, line 781: One thing I noticed testing this with real sgpmetE13.b1 data is that circular variables such as wdir_vec_mean will receive the default bin_average unless explicitly overridden. In my test, arithmetic bin averaging differed from a circular mean by >20° in 311 30-minute bins, with a worst-case difference of 180° (174.9° vs 354.9°). Could we document circular variables as another case requiring special handling, or consider supporting a circular-mean transform?

AdamTheisen (commented):
Overall, this looks to be a great addition to ACT that will help with ARM's datasets.  I did some testing and came across a few things but overall the results  were looking good and emphasized why using actual bounds matter.  Let us know what you think on these comments.

